PR #3, opened 2023-07-13
tests: Add suite for VM nesting scenarios

roosterfish:
Follow up on https://github.com/canonical/lxd/pull/11993

### DIFF
--- a/tests/vm_nesting.sh
+++ b/tests/vm_nesting.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+
+snap_channel="latest/edge"
+
+function parallel() {
+        seq 1 $1 | xargs -P $1 -I "{}" "${@:2}"
+}
+
+function init() {
+	parallel $1 lxc init ubuntu:jammy "t{}" $2
+}
+
+function conf() {
+	parallel $1 lxc config set "t{}" $2
+}
+
+function device_add() {
+	parallel $1 lxc config device add "t{}" $2 $3 $4
+}
+
+function start() {
+	args=""
+
+	for i in $(seq 1 $1); do
+		args="t$i $args"
+	done
+
+	echo "Start $args"
+	lxc start $args
+}
+
+function wait() {
+	parallel $1 bash -c "while true; do if lxc shell t{}; then break; fi; sleep 1; done"
+}
+
+function copy() {
+	parallel $1 lxc file push $2 "t{}$3"
+}
+
+function cmd() {
+	parallel $1 lxc exec "t{}" -- bash -c "$2"
+}
+
+function delete() {
+	args=""
+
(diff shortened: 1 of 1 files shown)


tests/vm_nesting.sh
```diff
@@ -0,0 +1,95 @@
...
+init 10 --vm
+conf 10 volatile.vsock_id=42
+start 10
+delete 10
+
+# Test 5 VMs each with one nested VM
+init 5 --vm
+start 5
+wait 5
+cmd 5 "snap wait system seed.loaded && snap refresh lxd --channel latest/stable"
```
tomponline, line 74: Can these tests always use latest/edge so we can test pre-release versions too

tests/vm_nesting.sh
```diff
@@ -0,0 +1,95 @@
...
+	echo "delete $args"
+	lxc delete -f $args
+}
+
+if [[ ! -f $1 ]]
+then
+	echo "Specify the path to the LXD binary for sideloading"
+	exit 1
+fi
+
```
tomponline, line 58: Can you add a snap install lxd --channel=latest/edge here so that it can be run on a fresh vm

tests/vm_nesting.sh
roosterfish: I have added a variable to select the edge channel. If `lxc` isn't already present on the host the test runs on, it gets installed. I haven't added a `snap refresh` so that the test doesn't modify our test environments if executed on the dev machines.

tomponline: You should be running latest/edge on dev machines though ;)

roosterfish: I have the bleeding edge always sideloaded :)

tomponline: Yes, but you need to also be running the latest/edge channel so that your sideloaded binary is being run in the latest snap environment and with latest dependencies (such as qemu):

https://github.com/canonical/lxd-pkg-snap/commits/latest-edge

roosterfish: Ah yes, thanks for the hint! Just changed it.

tests/vm_nesting.sh
```diff
@@ -0,0 +1,102 @@
...
+if [[ ! $(which lxc) ]]
```
simondeziel, line 61: Please use `command -v` instead of `which`.

roosterfish: Didn't know about `command`, changed.

simondeziel: It's a shell builtin and at one point `which` in `debianutils` was deprecated and complaining about it. So `command -v` is faster and should be trouble free :)

roosterfish: Thanks for the insight, I am all in for stability :)

PR #4, opened 2023-07-20
Adapt some tests scripts from lxc-ci to Openstack

simondeziel:
From the `main.sh` test list, `cgroup1` is failing on `jammy`. Previously in `lxc-ci` we used to test `cgroup2` along with `jammy` but I think that's already `jammy`'s default which is why I switched to `cgroup1`:

```
./bin/openstack-run jammy cgroup1 tests/cgroup
```

However, the following works:

```
./bin/openstack-run focal default tests/cgroup
```

So it seems like a regression to me.


tomponline (changes requested):
Looks like this duplicates some of the tests I cleaned up (partially} today and pushed in as part of the lxd 5.16 manual test run. Please can you update it so there's n9 duplication. Thanks

cloud-init.user-data.yaml
```diff
@@ -0,0 +1,32 @@
+#cloud-config
+apt:
+  # no need for backports/security nor restricted/multiverse
```
tomponline, line 3: Could you explain why in the comments as well as what.

tomponline (commented):
Looking good! Thanks

simondeziel:
I couldn't find the duplicated manual tests you referred to, would you mind pointing them out, please?

tomponline:
> 

https://github.com/canonical/lxd-ci/blob/main/tests/test-lxd-network-bridge-firewall
https://github.com/canonical/lxd-ci/blob/main/tests/test-lxd-network-ovn
https://github.com/canonical/lxd-ci/blob/main/tests/test-lxd-storage-buckets

I missed interception actually so would be good to bring that one across.

tomponline (approved):

PR #6, opened 2023-08-01
tests/interception: test security.syscalls.intercept.bpf if the kernel is new enough

simondeziel:


simondeziel:
I tested that PR on:
* focal 5.4 kernel: skips the test properly
* focal 5.15 kernel: works
* jammy 5.15 kernel: works

Prior to adding the kernel version check to skip this on old kernel, here's what I'd get from focal/5.4:

```
+ lxc config set c1 security.syscalls.intercept.bpf=true security.syscalls.intercept.bpf.devices=true
+ lxc restart c1 -f
Error: Seccomp notify doesn't support adding file descriptors
Try `lxc info --show-log c1` for more info
```

On a related note: @tomponline, we discussed the interception testing this morning and I said `setfattr` wasn't working but I was mistaken, it behaves just fine.

PR #17, opened 2023-09-06
Add `gpu-container` test

simondeziel:
The old script was testing with 3 different GPUs but the test system I was provided only had 1 so I adapted the script accordingly.

@morphis if you want to test with multiple GPUs, please let me know. I picked the `525` drivers but should I use the `535` instead? Does it matter? What about the `-server` utils?


tomponline (approved):
Happy to merge this once confirmed tested and working on @morphis server

simondeziel:
From a private chat with @morphis:
> yeah that looks good, I will release the VM and then schedule some work to get this integrated into our Jenkins

tomponline (approved):

PR #19, opened 2023-09-18
Refresh `network-ovn` and make it work without IPv6

simondeziel:
This was successfully ran on PS5.

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -53,7 +53,7 @@ ovs-vsctl set open_vswitch . \
 lxc project switch default
 lxc storage create default zfs
 
-instanceImage="images:ubuntu/22.04"
+instanceImage="ubuntu:22.04"
 
 ovn_basic_tests() {
     lxc network create lxdbr0 \
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -43,6 +43,13 @@ lxd waitready --timeout=300
 # Install dependencies
 apt-get install --no-install-recommends --yes ovn-host ovn-central bind9-dnsutils jq
 
+# Check for IPv6 uplink connectivity
+IPV6_UPLINK="true"
+if ! ip -6 ro get :: >/dev/null 2>&1; then
+  echo "No IPv6 uplink connectivity" >&2
+  IPV6_UPLINK="false"
+fi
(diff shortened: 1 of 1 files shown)


tomponline (approved):
I think later we could rework these tests to use lxdbr0 as an uplink so that we can test ipv6 and DNS to the uplink even if the host itself doesn't have external ipv6 connectivity.

simondeziel:
> I think later we could rework these tests to use lxdbr0 as an uplink so that we can test ipv6 and DNS to the uplink even if the host itself doesn't have external ipv6 connectivity.

In several places, `lxdbr0`'s IPv4/6 are tested like this:

```
ovn_basic_tests() {
    lxc network create lxdbr0 \
        ipv4.address=10.10.10.1/24 ipv4.nat=true \
        ipv4.dhcp.ranges=10.10.10.2-10.10.10.199 \
        ipv4.ovn.ranges=10.10.10.200-10.10.10.254 \
        ipv6.address=fd42:4242:4242:1010::1/64 ipv6.nat=true \
        ipv6.ovn.ranges=fd42:4242:4242:1010::200-fd42:4242:4242:1010::254
...

    echo "==> OVN to lxdbr0 gateway"
    lxc exec u2 -- ping -nc1 -w5 10.10.10.1
    lxc exec u2 -- ping -nc1 -w5 fd42:4242:4242:1010::1
```

Is this what you had in mind?

tomponline:
Yeah and then we can ping a host by name on lxdbr0 from ovn to test external DNS too

simondeziel:
I think all that is done already. I changed some `ping` to use TCP/443 probes due to PS5 outbound firewall but we already have that:

```
    echo "==> lxdbr0 to internet"
    lxc exec u1 -- nc -zv -4 -w5 ubuntu.com 443
    [ "${IPV6_UPLINK}" = "true" ] && lxc exec u1 -- nc -zv -6 -w5 ubuntu.com 443
```

So anything I am missing/misunderstood?

tomponline:
Yes, we dont need to ping externally at all really, all we need to do is check for connectivity to the uplink network.
So we don't need the IPV6_UPLINK var at all.

PR #20, opened 2023-09-19
tests: Improve network test reliability

tomponline:
This PR switches to using cloud-init with a `write_files` statement that sets up a per-boot script that sets the instance's state to READY via curl to /dev/lxd.

It then introduces a `waitInstanceReady` helper to wait until the instance reaches the known READY state.

### DIFF
--- a/tests/network
+++ b/tests/network
@@ -245,6 +245,7 @@ networkTests
 
 # Check VM "agent.nic_config" works by reconfiguring eth0 to use parent and mtu settings.
 echo "=> Performing VM bridged NIC agent.nic_config tests"
+lxc exec v1-physical -- "sync"
 lxc stop v1-physical -f
 lxc config set v1-physical agent.nic_config=true
 lxc config device set v1-physical eth0 nictype=bridged parent=lxdbr0 network= mtu=1400 name=eth0
--- a/tests/network
+++ b/tests/network
@@ -162,7 +162,7 @@ networkTests() {
         fi
 
         # DNS resolution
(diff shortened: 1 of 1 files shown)


tests/network
```diff
@@ -68,22 +68,17 @@ ip a
...
+cat <<EOF | lxc profile set default cloud-init.user-data -
+#cloud-config
+packages:
+  - curl
+write_files:
+  - content: |
+     #!/bin/bash
+     curl --unix-socket /dev/lxd/sock lxd/1.0 -X PATCH -d '{"state":"Ready"}'
+    path: /var/lib/cloud/scripts/per-boot/ready.sh
+    permissions: "0755"
```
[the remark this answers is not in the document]
tomponline: Yeah see https://cloudinit.readthedocs.io/en/latest/reference/modules.html#scripts-per-boot

This is different to `runcmd` which is only run on first boot.

markylaing (dismissed):
LGTM

tests/network
```diff
@@ -68,22 +68,17 @@ ip a
...
-    accept-ra: true
-    dhcp4: true
-    dhcp6: true
+cat <<EOF | lxc profile set default cloud-init.user-data -
+#cloud-config
+packages:
+  - curl
+write_files:
+  - content: |
+     #!/bin/bash
```
simondeziel, line 77: `#!/bin/sh`, please.

tests/network
```diff
@@ -68,22 +68,17 @@ ip a
...
+     curl --unix-socket /dev/lxd/sock lxd/1.0 -X PATCH -d '{"state":"Ready"}'
```
simondeziel, line 78: `exec curl`, please.

simondeziel (approved):

tests/network
tomponline: fixed thanks

simondeziel (approved):

PR #26, opened 2023-09-20
network-ovn: improve reliability

simondeziel:
It's still not perfect but seems to be a bit more reliable

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -215,7 +215,7 @@ ovn_basic_tests() {
     [ "${U3_IPV4_OLD}" = "${U3_IPV4}" ]
 
     echo "===> Testing project restrictions"
-    lxc project create testovn -c features.networks=true -c features.images=false -c restricted=true
+    lxc project create testovn -c features.networks=true -c features.images=false -c features.storage.volumes=false -c restricted=true
     lxc profile device add default root disk path=/ pool=default --project testovn
     setupCloudInitReady testovn
 
@@ -457,7 +457,7 @@ ovn_basic_tests() {
     ip link delete dummy0
 
     echo "===> Testing projects"
-    lxc project create testovn -c features.networks=true -c features.images=false -c limits.networks=1
+    lxc project create testovn -c features.networks=true -c features.images=false -c features.storage.volumes=false -c limits.networks=1
     lxc project switch testovn
     lxc profile device add default root disk path=/ pool=default
     setupCloudInitReady testovn
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -28,6 +28,7 @@ waitInstanceReady() (
   for i in $(seq 90) # Wait up to 90s.
   do
     if lxc info --project "${instProj}" "${instName}" | grep -qF "Status: READY"; then
+      sleep 5 # Allow some time for DHCP/SLAAC
       return 0 # Success.
     fi
 
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -94,9 +94,32 @@ ovs-vsctl set open_vswitch . \
 # Configure LXD.
 lxc project switch default
 lxc storage create default zfs
+lxc storage volume create default images
+lxc config set storage.images_volume=default/images # Speeds up instance creation.
 setupCloudInitReady default
 
-instanceImage="ubuntu:22.04"
+setupInstanceImage() {
+  lxc image copy ubuntu-minimal:22.04 local: --alias jammy-minimal
+  lxc init jammy-minimal jammy-small-tmp -s default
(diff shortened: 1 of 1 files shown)


tests/network-ovn
```diff
@@ -19,16 +19,21 @@ waitInstanceReady() (
   set +x
   # shellcheck disable=SC3043
   local instName="$1"
+  # shellcheck disable=SC3043
+  local instProj="${2:-}"
+  if [ -z "${instProj}" ]; then
```
tomponline, line 24: What does this bit do?

tomponline: Does it look for current project? If so why don't we just drop the --project flag if no project arg ?

tests/network-ovn
```diff
@@ -27,6 +27,7 @@ waitInstanceReady() (
   for i in $(seq 90) # Wait up to 90s.
   do
     if lxc info --project "${instProj}" "${instName}" | grep -qF "Status: READY"; then
+      sleep 5 # Allow some time for DHCP/SLAAC
```
tomponline, line 30: This should be done by then because systemd wait online has occurred?

tests/network-ovn
```diff
@@ -38,6 +38,23 @@ waitInstanceReady() (
   return 1 # Failed.
 )
 
+cloudInitReady() {
```
tomponline, line 41: Let's call this setupCloudInitReady, as the former sounds like a statement of fact.

tomponline: We should probably add the profile name as an arg as well as project.

tests/network-ovn
simondeziel: This allows to report the project's name in case of timeout. I could update the error message to mention the instance in the currently selected project could get ready?

tomponline: ah ok, no thats fine, but a comment explaining why its there will suffice, thanks!

tomponline:
@simondeziel as our usage of lxd-ci has changed to (mostly if not exclusively) being checking out the repo on the target machines, I wouldn't be adverse these days to having some common functions we could source into each script.

simondeziel:
@tomponline yeah, pooling common functions into a imported script makes sense. I reserve that work for when I'll implement a `bin/lxd-run` script that will include the boilerplate needed to test in an ephemeral VM on a local (or remote?) LXD.

simondeziel:
@tomponline the switch to a customized image based on `ubuntu-minimal:22.04` is what finally made that test 100% reliable for me. I've been testing exclusively on PS5 so maybe it was better on your own local machine.

tomponline:
What was the difference between two images?

tests/network-ovn
```diff
@@ -48,8 +48,6 @@ setupCloudInitReady() {
 
   cat <<EOF | lxc profile set "${profile}" --project "${instProj}" cloud-init.user-data -
 #cloud-config
-packages:
-  - curl
```
tomponline: I originally needed this on the `network` scripts because the sriov part of the script needs `images:ubuntu/22.04` so it has the relevant intel nic drivers to work.

tests/network-ovn
```diff
@@ -215,7 +215,7 @@ ovn_basic_tests() {
     [ "${U3_IPV4_OLD}" = "${U3_IPV4}" ]
 
     echo "===> Testing project restrictions"
-    lxc project create testovn -c features.networks=true -c features.images=false -c restricted=true
+    lxc project create testovn -c features.networks=true -c features.images=false -c features.storage.volumes=false -c restricted=true
```
tomponline, line 241: what is this part for?

tests/network-ovn
simondeziel: With the move to `ubuntu:` and possibly `ubuntu-minimal:` we can assume `curl` to be present so not having it listed there saves the time to run `apt-get update`. 

Considering that tests would fail in an obvious way if for some reason `curl` would be missing, are you OK with keeping it that way?

tests/network-ovn
simondeziel: This was made in an attempt to figure out why `u1` was way slower to boot than any other instances. It's possibly not needed but it aligns with the other projects also having that same key/value set.

tests/network-ovn
tomponline: Yeah im fine with it being removed here, just a note that this approach wont work for all tests :)

simondeziel: Eventually, I'd like to migrate from `ubuntu:` to a local image based on a tweaked `ubuntu-minimal:` image, exactly as done in this PR. In such case, if a package would be missing we could decide to either bake it inside the base image or install it just for the test needing it.

PR #30, opened 2023-09-26
network-ovn: revert to using ubuntu:22.04

simondeziel:
The main issue was due to the `testovn` project using a dedicated set of profiles thus not having the cloud-init `curl Ready | /dev/lxd` bits applied from the default profile.

My use of a custom image based on `ubuntu-minimal:22.04` papered over this by forcing the `curl Ready | /dev/lxd` bits to be in the base image making it magically way faster :)

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -1558,7 +1558,7 @@ ovn_nested_vlan_tests() {
 
     waitInstanceReady u1
     waitInstanceReady u2
-    lxc ls
+    lxc list
     lxc exec u1 -- ip netns exec vlan_nesting1 ip address
     lxc exec u1 -- ip netns exec vlan_nesting2 ip address
 
@@ -1989,7 +1989,7 @@ ovn_l3only_tests() {
 
     waitInstanceReady u1
     waitInstanceReady u2
-    lxc ls
+    lxc list
 
     U1_IPV4="$(lxc list u1 -c4 --format=csv | cut -d' ' -f1)"
     U1_IPV6="$(lxc list u1 -c6 --format=csv | cut -d' ' -f1)"
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -29,7 +29,6 @@ waitInstanceReady() (
   fi
   for _ in $(seq "${maxWait}"); do
     if lxc info --project "${instProj}" "${instName}" | grep -qF "Status: READY"; then
-      sleep 5 # Allow some time for DHCP/SLAAC
       return 0 # Success.
     fi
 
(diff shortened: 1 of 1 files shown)


tests/network-ovn
```diff
@@ -99,7 +100,7 @@ lxc config set storage.images_volume=default/images # Speeds up instance creatio
 setupCloudInitReady default
 
 setupInstanceImage() {
-  lxc image copy ubuntu-minimal:22.04 local: --alias jammy-minimal
+  lxc image copy ubuntu:22.04 local: --alias jammy-minimal
```
tomponline: Shall we call the image alias just jammy?

tomponline: You don't really need this line, can just create the temporary instance directly from the remote image before publishing as jammy-small.

tests/network-ovn
```diff
@@ -2124,9 +2122,7 @@ else
     ovn_leases_tests
 fi
 
-lxc image delete "${instanceImage}"
-lxc config unset storage.images_volume
-lxc storage volume delete default images
+lxc profile device remove default root
```
simondeziel, line 2125: I don't understand why removing the `root` device from the default profile wasn't needed before? @tomponline any idea?

simondeziel:
Successful run on PS5:
```
+ lxc delete -f c1
+ lxc network delete ovn0
Network ovn0 deleted
+ lxc network delete lxdbr0
Network lxdbr0 deleted
+ lxc profile device remove default root
Device root removed from default
+ lxc storage delete default
Storage pool default deleted

Test passed
+ FAIL=0
+ cleanup
+ echo 
+ [ 0 = 1 ]
+ echo Test passed
+ exit 0
+ RET=0
+ cleanup
+ set +e
+ openstack server delete lxd-ci-network-ovn-jammy-638649
+ [ 0 = 0 ]
+ echo 

+ echo ==> Test passed (network-ovn)
==> Test passed (network-ovn)
+ exit 0

real    31m29.561s
user    0m13.996s
sys     0m2.791s
```

tests/network-ovn
tomponline: Before your recent set of changes we werent adding the root disk device to the profile but instead creating the instances directly using the `-s` flag to add a root disk to each instance directly so there was less clean up needed.

simondeziel: The root disk device as long been in the default project (since commit 88252e0d). What I did change is the `features.profiles` on some projects but it's still unclear to me what I introduced that needed that additional cleanup?

Looking at `git diff --follow  -p 88252e0d.. -- tests/network-ovn`, I'm not seeing anything that could explain this. Anyway, not something important but I was (still am) curious. Thanks!

PR #40, opened 2023-12-04
Testflinger alt kernel

simondeziel:


bin/custom-kernel
```diff
@@ -115,4 +115,11 @@ case "$i" in
     esac
 done
 
+# Since we are doing a reboot, might as well also turn off mitigations and have a faster testbed
+echo "===> Disabling all mitigations"
+
+# shellcheck disable=SC2016
+echo 'GRUB_CMDLINE_LINUX="${GRUB_CMDLINE_LINUX} mitigations=off"' > /etc/default/grub.d/99-mitigations.cfg
```
tomponline, line 120: Does this represent a realistic test environment though?

tomponline (approved):

bin/custom-kernel
simondeziel: That's a fair point however mitigations and their performance impact vary a lot from a system to another. For example, an AMD CPU won't have kPTI enabled while an Intel one would.

I'll do it differently in another PR where I will require `nomitigations` as the kernel option to have those mitigations disabled. This would add another row to the matrix of tests but might be interesting for the performance tests to compare with mitigations enabled.

PR #42, opened 2023-12-11
Assortement of fixes to better test `5.0/edge` and "exotic" arches

simondeziel:
The `waitInstanceReady` function was changed to work with `5.0/edge` that doesn't support instance self-reporting their readiness (`curl --unix-socket /dev/lxd/sock lxd/1.0 -X PATCH -d '{"state":"Ready"}'`). The new method relies on `systemctl is-system-ready --wait` which completes at the same time as `cloud-init status --wait` anyway but has the advantage of being more "portable".


tomponline:
> The `waitInstanceReady` function was changed to work with `5.0/edge` that doesn't support instance self-reporting their readiness (`curl --unix-socket /dev/lxd/sock lxd/1.0 -X PATCH -d '{"state":"Ready"}'`)

Does this catch when the lxd-agent is ready too?

tests/devlxd-vm
```diff
@@ -87,7 +87,9 @@ lxc start v1
 waitVMAgent v1
 
 # Configure to use the proxy
-lxc exec v1 -- snap install lxd --channel="${LXD_SNAP_CHANNEL:-"latest/edge"}"
+lxc exec v1 -- snap wait system seed.loaded || true
```
tomponline, line 90: why `|| true` here?

simondeziel:
> > The `waitInstanceReady` function was changed to work with `5.0/edge` that doesn't support instance self-reporting their readiness (`curl --unix-socket /dev/lxd/sock lxd/1.0 -X PATCH -d '{"state":"Ready"}'`)
> 
> Does this catch when the lxd-agent is ready too?

Yes as this comes as a systemd unit. `systemctl is-system-ready --wait` will return `running` if all units started without issue (including `lxd-agent.service`) but if some failures happen, it will return `degraded` so would cause `waitInstanceReady` to return an error.

tests/storage-buckets
```diff
@@ -4,6 +4,13 @@ set -eux
 # Install LXD
 install_lxd
 
+if ! hasNeededAPIExtension storage_buckets; then
```
tomponline, line 7: Could we add this skip logic to the main script that runs each test? 
This way if you manually run the test against an incompatible host you dont get a zero exit status.

tests/devlxd-vm
simondeziel: I'm afraid I don't remember, I'll drop it.

tomponline (approved):

tests/storage-buckets
simondeziel: Moving it to outside of the test script itself would make it impossible to do the API detection, no? In the main script, we only know about the snap channel, not the API extensions that come with it.

simondeziel: In the long run, when I'll have `bin/lxd-run` ready, I intend to include a testrunner image cache that will allow dumping the api extentions from the testrunner prior to "forking" it to run the tests. This will permit API detection ahead of test runs but are not there yet :/

PR #50, opened 2023-12-14
Make `tests/gpu-container` adapt to available HW

simondeziel:
This was tested on `rockman` using:

```
PURGE_LXD=1 ./bin/local-run tests/gpu-container latest/edge
```

`rockman` has the particularity of having 3 GPUs, one onboard Matrox and 2x NVIDIA L40.

### DIFF
--- a/tests/gpu-container
+++ b/tests/gpu-container
@@ -35,6 +35,18 @@ lxc profile device add default root disk path=/ pool=default
 lxc network create lxdbr0
 lxc profile device add default eth0 nic network=lxdbr0 name=eth0
 
+# Consult available resources
+total_gpu="$(lxc query /1.0/resources | jq -r '.gpu.total')"
+total_nvidia_gpu="$(lxc query /1.0/resources | jq -r '.gpu.cards | .[] | select(.driver == "nvidia") | .pci_address' | wc -l)"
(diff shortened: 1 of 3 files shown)


simondeziel:
@escabo with this, you should be able to run the `tests/gpu-container` unaltered on your own hardware, at least that's the theory and if that still requires modification to pass, I'd like to hear about it, thanks!

tests/gpu-container
```diff
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -eu
 
+# testflinger_queue: rockman
```
tomponline, line 4: This is fine for now, but eventually we'll want to move this out of the script itself and into the configuration file that schedules the run.

tomponline (approved):

tests/gpu-container
simondeziel: Indeed, that's the plan.

escabo:
> @escabo with this, you should be able to run the `tests/gpu-container` unaltered on your own hardware, at least that's the theory and if that still requires modification to pass, I'd like to hear about it, thanks!

I ran ```sudo bin/local-run tests/gpu-container latest/candidate``` and it failed.

The machine has secure boot and when I install the (Nvidia) driver manually, I have to go through creating a pw that I am asked on the next boot to register the new keys. Can/does the script work around this?

Also, the script installs 525 but if I run ```sudo ubuntu-drivers devices```, the recommended one is 535. Could this be part of the script to avoid having to update the required version every so often?

If I install the driver manually, the test works! I had to install jq on the host.

PR #51, opened 2023-12-17
Improve `tests/gpu-container` to pick the recommended driver series and abort on secureboot enabled

simondeziel:
Based on comments/suggestion from https://github.com/canonical/lxd-ci/pull/50#issuecomment-1858451719

@escabo, this addresses your feedback  however I couldn't find a nice way to avoid DKMS triggering the onboarding of custom MOKs and thus prompting for password. The test will warn and bail if secureboot is enabled.

@tomponline this was tested successfully on `rockman`.

### DIFF
--- a/tests/gpu-container
+++ b/tests/gpu-container
@@ -26,7 +26,9 @@ if mokutil --sb-state | grep -Fx "SecureBoot enabled"; then
 fi
 
 # Install dependencies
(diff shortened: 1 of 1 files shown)


simondeziel:
@tomponline please take a look when the time permits, thanks!

PR #55, opened 2024-01-12
Enable `tests/cluster` test

simondeziel:
This should be merged after #54.

This was tested successfully with:

```
PURGE_LXD=1 ./bin/local-run tests/cluster 5.0/edge 3 5.0/stable 5.0/edge
PURGE_LXD=1 ./bin/local-run tests/cluster latest/edge 3 latest/stable latest/edge
```

So upgrading `5.0/stable` -> `5.0/edge` and `latest/stable` -> `latest/edge`.

### DIFF
--- a/tests/network-sriov
+++ b/tests/network-sriov
@@ -134,21 +134,21 @@ lxc delete -f v2 v3 v4
 
 # Launch a few containers.
 echo "==> Container on default VLAN"
-lxc init ubuntu-daily:22.04 c1
+lxc init "${TEST_IMG:-ubuntu-daily:22.04}" c1
 lxc start c1
 
 echo "==> Container on default VLAN with filtering"
-lxc init ubuntu-daily:22.04 c2
+lxc init "${TEST_IMG:-ubuntu-daily:22.04}" c2
 lxc config device override c2 eth0 security.mac_filtering=true
 lxc start c2
 
 echo "==> Container on alternate VLAN"
-lxc init ubuntu-daily:22.04 c3
+lxc init "${TEST_IMG:-ubuntu-daily:22.04}" c3
 lxc config device override c3 eth0 vlan=4000
 lxc start c3
 
 echo "==> Container on alternate VLAN with filtering"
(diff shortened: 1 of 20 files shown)


tests/network-sriov
```diff
@@ -65,8 +65,8 @@ networkTests() {
         if echo "${address}" | grep ":" -q; then
             echo "PASS: IPv6 address: ${name}"
         else
-            echo "FAIL: IPv6 address: ${name}"
-            FAIL=1
+            echo "WARN: IPv6 address: ${name}"
+            #FAIL=1
```
tomponline, line 69: Out of interest, why is this?

tests/network-sriov
```diff
@@ -101,21 +101,21 @@ lxc profile device add default eth0 nic nictype=sriov parent="${parentNIC}" name
 # Launch a few VMs.
 # Do this first before containers to ensure VF free search handles VFs unbound from host.
 echo "==> VM on default VLAN"
-lxc init ubuntu-daily:23.10 v1 --vm
+lxc init ubuntu-daily:24.04 v1 --vm
```
tomponline, line 104: @simondeziel a thought for another time, but it'd be nice to be able to change this for all tests with an env var, and then default to a specific version in a single place in the code.

tomponline (dismissed):

tomponline:
@simondeziel ready for a rebase

tests/network-sriov
simondeziel: I've never seen the need for this as it's easy to change (`sed 's/:22\.04/:24.04/'`) either globally or on a single `tests/` file, IMHO. However if that is of use to you, sure. You'd like just the image name to be changeable? Would `$testImg` be a good variable name?

tomponline: I suppose something like `export TEST_IMG=ubuntu-daily:24.04` would be good.

simondeziel:
@tomponline please take another look.

tomponline (approved):

PR #61, opened 2024-02-05
Adapt VM storage tests to support Dell PowerFlex

roosterfish:
This PR adapts both of the VM storage tests so they can be executed using the Dell PowerFlex storage driver from https://github.com/canonical/lxd/pull/12304.

Storage size values for PowerFlex need to be in multiples of 8GiB. The minimal storage volume size in PowerFlex is also 8GiB. 
Where applicable the storage size of the actual test was increased for all the drivers to accommodate PowerFlex. At other places a different storage size gets set for the PowerFlex driver only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,17 @@ Or to run individual tests (`tests/pylxd` against `latest/edge`):
 ./bin/openstack-run jammy default tests/pylxd latest/edge
 ```
 
+# Running Dell PowerFlex VM storage tests
+
+To run the VM storage tests on the Dell PowerFlex driver, provide the following environment variables:
+
+* `POWERFLEX_POOL`: Name of the PowerFlex storage pool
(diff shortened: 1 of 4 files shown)


simondeziel (commented):
I'm fine with the changes but the `tests/storage-vm` goes from ~86 minutes to 115 minutes on PS5 instances. If we care about testing this on OpenStack instances, we might want to tweak the disk sizes like done in `tests/storage-volumes-vm`.

simondeziel (dismissed):
Other than the slower speed of `tests/storage-vm`, things still work as tested on non-powerflex storage driver.

tomponline:
Can we make the larger sizes only be used for the powerflex tests?

simondeziel (approved):
Thanks!

tomponline (approved):

PR #66, opened 2024-02-09
Install and remove MinIO for storage buckets tests

masnax:
As MinIO is no longer bundled with LXD, we have to build it manually for the tests and set `minio.path` on the LXD snap.

### DIFF
--- a/tests/storage-buckets
+++ b/tests/storage-buckets
@@ -11,6 +11,21 @@ if ! hasNeededAPIExtension storage_buckets; then
   exit 0
 fi
 
+
+# Grab the architecture so we can install the correct pre-built minio.
+arch="$(uname -m)"
+if [ "${arch}" = "aarch64" ]; then
+  arch="arm64"
+elif [ "${arch}" = "ppc64el" ]; then
+  arch="ppc64le"
+elif [ "${arch}" = "armhf" ]; then
+  arch="arm"
+elif [ "${arch}" = "riscv64" ] ; then
+  echo "Skipping test as not supported by architecture $(arch)"
+  FAIL=0
+  exit 0
+fi
+
 poolDriverList="${1:-dir btrfs lvm lvm-thin zfs ceph}"
 
 if echo "${poolDriverList}" | grep -qwF "ceph"; then
@@ -21,6 +36,23 @@ if echo "${poolDriverList}" | grep -qwF "ceph"; then
   apt-get update
   apt-get install --no-install-recommends --yes ceph-common
 fi
+
+# Clean up the build dir in case it hung around from a failed test.
+rm -rf /opt/minio
+mkdir -p /opt/minio
+
+# Download the minio server.
+curl -sSfL "https://dl.min.io/server/minio/release/linux-${arch}/minio" --output "/opt/minio/minio"
+chmod +x "opt/minio/minio"
+
+# Also grab the latest minio client to maintain compatibility with the server.
+curl -sSfL "https://dl.min.io/client/mc/release/linux-${arch}/mc" --output "/opt/minio/mc"
+chmod +x "opt/minio/mc"
+
+# Set the snap config key for minio and reload LXD to have it take effect.
+snap set lxd minio.path=/opt/minio
+systemctl reload snap.lxd.daemon
+lxd waitready --timeout=300
 
 # Configure LXD
 lxc project switch default
@@ -60,6 +92,9 @@ done
 
 lxc project switch default
 lxc project delete test
+
+# Remove MinIO now that the test is over.
+rm -rf /opt/minio
 
 # shellcheck disable=SC2034
 FAIL=0


masnax:
@tomponline @simondeziel the snap bundled MinIO pinned to a specific version, I just used the latest release here, but if it's preferable to keep the old pinned version then I can use that instead.

tests/storage-buckets
```diff
@@ -22,6 +27,20 @@ if echo "${poolDriverList}" | grep -qwF "ceph"; then
   apt-get install --no-install-recommends --yes ceph-common
 fi
 
+# Fetch MinIO from its latest GitHub release tarball.
+minio_address="$(curl -s "https://api.github.com/repos/minio/minio/releases/latest" | jq '.tarball_url' | sed 's/"//g')"
```
simondeziel, line 31: `jq -r` will avoid needing `sed`.

tests/storage-buckets
```diff
@@ -22,6 +27,20 @@ if echo "${poolDriverList}" | grep -qwF "ceph"; then
...
+curl -L "${minio_address}" -o /tmp/minio.tar.gz
+
+# Clean up the build dir in case it hung around from a failed test.
+rm -rf /tmp/minio
+mkdir /tmp/minio
```
simondeziel, line 36: I would avoid `/tmp` even with a test/CI script as using hardcoded path in there as root is insecure.

tests/storage-buckets
```diff
@@ -22,6 +27,20 @@ if echo "${poolDriverList}" | grep -qwF "ceph"; then
...
+
+# Build MinIO and apply it to the LXD snap.
+tar -xf /tmp/minio.tar.gz -C /tmp/minio
+make build -C /tmp/minio/*/
+mv /tmp/minio/*/minio /tmp/minio/
```
simondeziel, line 41: Upstream provides builds for many arches according to https://min.io/download#/linux. If we can avoid building that'd be even better.
```
wget https://dl.min.io/server/minio/release/linux-amd64/minio
chmod +x minio
```

We can likely rely on `uname -m` with some mangling for `aarch64` and `ppc64el` mangling as usual but aside from that we should be good. They even provide `s390x`!

tests/storage-buckets
simondeziel: `/opt/minio` feels like a better place it requires root privs to write into.

tests/storage-buckets
masnax: Ah nice, very convenient. How come we were building it in the snap though?

simondeziel: I don't know but maybe because we never used built artifacts except for those coming through Ubuntu packages? Also, it isn't very nice from a reproducible point of view to always pick their latest release but maybe they provide older versions somewhere too?

masnax: Yeah, I guess if it always pulls the latest version then a `stable` snap would never actually be stable.

tomponline: Our GH pulls the prebuilt asset so for testing this is fine to do

tomponline:
> @tomponline @simondeziel the snap bundled MinIO pinned to a specific version, I just used the latest release here, but if it's preferable to keep the old pinned version then I can use that instead.
> 
> Might look a bit cleaner that way too, since we could just `git clone` and know which tag we want.

Using latest is fine for tests

tests/storage-buckets
```diff
@@ -22,6 +36,23 @@ if echo "${poolDriverList}" | grep -qwF "ceph"; then
...
+curl -sSfL "https://dl.min.io/server/minio/release/linux-${arch}/minio" --output "/opt/minio/minio"
+chmod +x "opt/minio/minio"
+
+# Also grab the latest minio client to maintain compatibility with the server.
+curl -sSfL "https://dl.min.io/client/mc/release/linux-${arch}/mc" --output "/opt/minio/mc"
+chmod +x "opt/minio/mc"
+
+# Set the snap config key for minio and restart LXD to have it take effect.
+snap set lxd minio.path=/opt/minio
+snap restart lxd
```
tomponline, line 53: Lxd reload should be sufficient and snap restart has a bug with lxd so let's not use it here

tests/storage-buckets
```diff
@@ -11,6 +11,20 @@ if ! hasNeededAPIExtension storage_buckets; then
...
+
+# Grab the architecture so we can install the correct pre-built minio.
+arch="$(uname -m)"
+if [ "${arch}" = "aarch64" ]; then
+  arch="arm64"
+elif [ "${arch}" = "ppc64el" ]; then
+  arch="ppc64le"
+fi
+
+if ! echo "amd64 arm64 ppc64le s390x" | grep -q "${arch}" ; then
```
simondeziel, line 23:
```suggestion
if ! echo "amd64 arm64 ppc64le s390x" | grep -qwF "${arch}" ; then
```

tests/storage-buckets
```diff
@@ -11,6 +11,20 @@ if ! hasNeededAPIExtension storage_buckets; then
...
+  echo "Skipping test as not supported by architecture $(arch)"
```
simondeziel, line 24:
```suggestion
  echo "Skipping test as not supported by architecture $(arch)"
  FAIL=0
```

tests/storage-buckets
simondeziel, line 23: It seems like `armhf` is also supported https://dl.min.io/server/minio/release/linux-arm/
I think you'll also need some `uname -m` mangling here too though.

In the end, all our arches except for `riscv64` are supported so maybe you'd prefer to invert the checking logic.

tests/storage-buckets
```diff
@@ -22,6 +36,23 @@ if echo "${poolDriverList}" | grep -qwF "ceph"; then
...
+
+# Download the minio server.
+curl -sSfL "https://dl.min.io/server/minio/release/linux-${arch}/minio" --output "/opt/minio/minio"
+chmod +x "opt/minio/minio"
+
+# Also grab the latest minio client to maintain compatibility with the server.
+curl -sSfL "https://dl.min.io/client/mc/release/linux-${arch}/mc" --output "/opt/minio/mc"
+chmod +x "opt/minio/mc"
+
+# Set the snap config key for minio and restart LXD to have it take effect.
```
simondeziel, line 51:
```suggestion
# Set the snap config key for minio and reload LXD to have it take effect.
```

simondeziel (commented):
This looks good for `latest/edge` but the same tests need to work on `5.0/edge` where it'd be ideal not to download those bits.

tests/storage-buckets
simondeziel: Even when we vendored `minio`, `riscv64` was not supported.

masnax:
> This looks good for `latest/edge` but the same tests need to work on `5.0/edge` where it'd be ideal not to download those bits.



> This looks good for `latest/edge` but the same tests need to work on `5.0/edge` where it'd be ideal not to download those bits.

This got me a bit confused. So is minio not going to be supported on 5.0/edge? If it is, then will it continue to be bundled in the snap? If not, then how else would we fetch the binaries if we don't download them?

simondeziel:
@masnax minio will remain in `5.0/*`. So ideally we would download the `minio`/`mc` binaries only if needed.

tests/storage-buckets
masnax: I'm not sure where to find "our architctures" so I just tried grepping the codebase and I see some references to `ppc64` and `mips64le` which aren't supported by minio (though they do support `ppc64le` and `mips64`). Do we run tests for these?

simondeziel: https://launchpad.net/~canonical-lxd/+snap/lxd-latest-edge/+edit show you the list of arches for which we build the snap for at the bottom

masnax: Thanks!

simondeziel:
Hmm, https://github.com/canonical/lxd-pkg-snap/blob/5.0-candidate/snapcraft.yaml has no `minio` part at all. I'm confused, did we even shipped 5.0 with minio bundled or was that just a feature release thing?

simondeziel:
@masnax so Tom confirmed, we never shipped the minio bundled server in `5.0/*`, only the client bits. That said, the script you are modifying bails if `storage_buckets` is not a supported extension so please disregard all I said about 5.0 as this is already bypassed thanks to that API support check.

simondeziel (approved):

PR #68, opened 2024-02-12
tests/efi-vars-editor-vm: add a new test

mihalicyn:
Depends on:
https://github.com/canonical/lxd/pull/12870

### DIFF
--- a/tests/efi-vars-editor-vm
+++ b/tests/efi-vars-editor-vm
@@ -1,0 +1,89 @@
+#!/bin/sh
+set -eux
+
+architecture="$(uname -m)"
+if [ "${architecture}" != "x86_64" ] && [ "${architecture}" != "aarch64" ]; then
+  echo "Skipping test as EFI is not supported on ${architecture}"
+  # shellcheck disable=SC2034
+  FAIL=0
+  exit 0
+fi
+
+# Install LXD
+install_lxd
+
+if ! hasNeededAPIExtension instances_uefi_vars; then
+  echo "Skipping test as not supported"
+  # shellcheck disable=SC2034
+  FAIL=0
+  exit 0
+fi
+
+# Configure LXD
+lxc network create lxdbr0
+lxc profile device add default eth0 nic network=lxdbr0
+
+poolName="vmpool$$"
+poolDriver=dir
+
+echo "==> Create storage pool using driver ${poolDriver}"
+lxc storage create "${poolName}" "${poolDriver}"
+
+extra_cleanup() {
+  echo "==> extra_cleanup"
+
+  echo "==> Stopping and deleting VM"
+  lxc delete -f v1
+
+  lxc profile device remove default eth0
+
+  echo "==> Deleting storage pool"
+  lxc storage delete "${poolName}"
+
+  echo "==> Deleting network"
+  lxc network delete lxdbr0
+}
+
+echo "==> Create VM and boot"
+lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm -s "${poolName}"
+lxc info v1
+
+EFI_GLOBAL_VARIABLE_GUID="8be4df61-93ca-11d2-aa0d-00e098032b8c"
+EFI_VAR_NAME="testvar"
+EFI_VAR_FULL_NAME="${EFI_VAR_NAME}-${EFI_GLOBAL_VARIABLE_GUID}"
+
+# "Hello from LXD" in ASCII
+EFI_VAR_VALUE_ASCII="Hello from LXD"
+EFI_VAR_VALUE="48656c6c6f2066726f6d204c5844"
+
+echo "==> Try to set EFI variable while VM is running (should fail)"
+! lxc config uefi set v1 "${EFI_VAR_FULL_NAME}"="${EFI_VAR_VALUE}" || false
+
+# Stop VM and set EFI variable
+lxc stop -f v1
+
+echo "==> Try to set invalid data as EFI variable value"
+# Try to set invalid data ("0" is not a valid HEX-encoding for a byte, should be "00")
+! lxc config uefi set v1 "${EFI_VAR_FULL_NAME}"="0" || false
+
+echo "==> Set a valid variable data"
+lxc config uefi set v1 "${EFI_VAR_FULL_NAME}"="${EFI_VAR_VALUE}"
+
+echo "==> Start VM and try to read a set value back from efivarfs"
+lxc start v1
+waitInstanceReady v1
+lxc exec v1 -- grep -xF "${EFI_VAR_VALUE_ASCII}" /sys/firmware/efi/efivars/"${EFI_VAR_FULL_NAME}"
+
+echo "==> Try to unset with VM running (should fail)"
+! lxc config uefi unset v1 "${EFI_VAR_FULL_NAME}" || false
+
+lxc stop -f v1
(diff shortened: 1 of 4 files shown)


tests/efi-vars-editor-vm
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
+set -eux
+
+architecture="$(uname -m)"
+if [ "${architecture}" != "x86_64" ] && [ "${architecture}" != "aarch64" ]; then
+  echo "Skipping test as EFI is not supported on ${architecture}"
```
simondeziel, line 6: This would need the following to stop here:
```
FAIL=0
exit 0
```

Is the `lxc config uefi` support not advertised via an API extension? Or is that advertised for all arches because API extensions are generic?

tests/efi-vars-editor-vm
```diff
@@ -0,0 +1,73 @@
...
+lxc profile device add default eth0 nic network=lxdbr0
+
+poolName="vmpool$$"
+poolDriver=dir
+
+echo "==> Create storage pool using driver ${poolDriver}"
+lxc storage create "${poolName}" "${poolDriver}"
+
+echo "==> Create VM and boot"
+lxc launch "${TEST_IMG:-ubuntu-daily:22.04}" v1 --vm -s "${poolName}"
```
simondeziel, line 23: Pleas use `ubuntu-minimal-daily:22.04" here, it's much faster on slow CI.

tests/efi-vars-editor-vm
```diff
@@ -0,0 +1,73 @@
...
+# Try to set invalid data ("0" is not a valid HEX-encoding for a byte, should be "00")
+! lxc config uefi set v1 "${EFI_VAR_FULL_NAME}"="0" || false
+
+# Set a valid variable data
+lxc config uefi set v1 "${EFI_VAR_FULL_NAME}"="${EFI_VAR_VALUE}"
+
+# Start VM and try to read a set value back from efivarfs
+lxc start v1
+waitInstanceReady v1
+lxc exec v1 -- cat /sys/firmware/efi/efivars/"${EFI_VAR_FULL_NAME}" | grep "${EFI_VAR_VALUE_ASCII}"
```
simondeziel, line 50: Useless use of `cat`.
```suggestion
lxc exec v1 -- grep -xF "${EFI_VAR_VALUE_ASCII}" /sys/firmware/efi/efivars/"${EFI_VAR_FULL_NAME}"
```
This also uses a more strict `grep` pattern, assuming that value is really the whole line we expect.

tests/efi-vars-editor-vm
```diff
@@ -0,0 +1,73 @@
...
+
+# Stop VM and unset EFI variable
+lxc stop -f v1
+lxc config uefi unset v1 "${EFI_VAR_FULL_NAME}"
```
simondeziel, line 54: Can we unset while running? Might want to test that too.

tests/efi-vars-editor-vm
```diff
@@ -0,0 +1,73 @@
...
+echo "==> Create VM and boot"
+lxc launch "${TEST_IMG:-ubuntu-daily:22.04}" v1 --vm -s "${poolName}"
+waitInstanceReady v1
+lxc info v1
+
+EFI_GLOBAL_VARIABLE_GUID="8be4df61-93ca-11d2-aa0d-00e098032b8c"
+EFI_VAR_NAME="testvar"
+EFI_VAR_FULL_NAME="${EFI_VAR_NAME}-${EFI_GLOBAL_VARIABLE_GUID}"
+
+# "Hello from LXD" in ASCII
```
simondeziel, line 31: Please turn those in to `echo ==>` lines as it eases navigating through CI logs when debugging :)

tests/efi-vars-editor-vm
```diff
@@ -0,0 +1,73 @@
...
+lxc exec v1 -- cat /sys/firmware/efi/efivars/"${EFI_VAR_FULL_NAME}" | grep "${EFI_VAR_VALUE_ASCII}"
+
+# Stop VM and unset EFI variable
+lxc stop -f v1
+lxc config uefi unset v1 "${EFI_VAR_FULL_NAME}"
+
+# Start VM and check that variable does not exist anymore
+lxc start v1
+waitInstanceReady v1
+! lxc exec v1 -- cat /sys/firmware/efi/efivars/"${EFI_VAR_FULL_NAME}" || false
```
simondeziel, line 59: I'd probably use `test -e` instead as that is clearer what the intentions are.

tests/efi-vars-editor-vm
```diff
@@ -0,0 +1,73 @@
...
+echo "==> Stopping and deleting VM"
+lxc delete -f v1
+
+lxc profile device remove default eth0
+
+echo "==> Deleting storage pool"
+lxc storage delete "${poolName}"
+
+echo "==> Deleting storage pool"
+lxc network delete lxdbr0
```
simondeziel, line 70: Please put that in a `extra_cleanup` function that is skipped on failure in CI, saving every CI minutes we can :)

tests/efi-vars-editor-vm
mihalicyn: It's advertised for all arches, yes, because API extensions are generic.

And yes, this is definitely a copy-n-paste mistake, because I was looking on `tests/cpu-vm` as an example. It has the same issue.

tests/efi-vars-editor-vm
mihalicyn: Sure. Thanks!

tests/efi-vars-editor-vm
mihalicyn: No, we can only read while running. Will add this too. Thanks!

tests/efi-vars-editor-vm
mihalicyn: sure, my bad. Will fix.

tests/efi-vars-editor-vm
mihalicyn: Thanks, fixed!

tests/efi-vars-editor-vm
mihalicyn: Have fixed!

tests/efi-vars-editor-vm
```diff
@@ -0,0 +1,83 @@
...
+  echo "==> Deleting storage pool"
+  lxc storage delete "${poolName}"
+
+  echo "==> Deleting network"
+  lxc network delete lxdbr0
+}
+
+echo "==> Create VM and boot"
+lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm -s "${poolName}"
+waitInstanceReady v1
```
simondeziel, line 42: Do we actually need for the boot to complete at this point?

mihalicyn: I don't think so. Have removed.

simondeziel (dismissed):
LGTM, thanks!

simondeziel:
@mihalicyn please include this new test in here: https://github.com/canonical/lxd-ci/blob/main/.github/workflows/tests.yml#L60

simondeziel (dismissed):

tomponline (commented):
Do we need to add API extension support to this test?

tomponline:
> Have added check for API extension. Thanks!
> 
> Failures looks non related to my change. Something is wrong with OVN...

Yeah OVN and the virtiofsd issues are known and @roosterfish has already landed a fix for virtiofsd.

tomponline (approved):

PR #69, opened 2024-02-14
Sync image definitions with lxc-ci

MusicDin:
Cherry picked from [lxc/lxc-ci](https://github.com/lxc/lxc-ci)

### DIFF
--- a/images/ubuntu.yaml
+++ b/images/ubuntu.yaml
@@ -662,13 +662,9 @@ actions:
 
     # Enable systemd-networkd
     systemctl enable systemd-networkd
-  releases:
-  - bionic
-  - focal
-  - jammy
-  - lunar
-  - mantic
-  - noble
+
+    # Disable UA attach
+    systemctl mask ua-auto-attach
 
 - trigger: post-packages
   action: |-
--- a/images/ubuntu.yaml
+++ b/images/ubuntu.yaml
@@ -343,7 +343,6 @@ files:
   - bionic
   - focal
   - jammy
-  - lunar
   - mantic
   - noble
   types:
@@ -384,7 +383,6 @@ files:
   - bionic
   - focal
   - jammy
-  - lunar
   - mantic
   - noble
   types:
@@ -489,7 +487,6 @@ packages:
     - fuse3
     releases:
     - jammy
-    - lunar
     - mantic
     - noble
     action: install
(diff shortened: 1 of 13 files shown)


tomponline:
@MusicDin were these tested as i note: https://github.com/lxc/lxc-ci/issues/813

simondeziel:
Considering the source and destination repos are both under Apache-2.0 license, I don't think we should carry the `License: Apache-2.0` on those cherry-picked commits. @tomponline what's your opinion on this?

tomponline:
> Considering the source and destination repos are both under Apache-2.0 license, I don't think we should carry the `License: Apache-2.0` on those cherry-picked commits. @tomponline what's your opinion on this?

I've got an outstanding question for the LXD management team as to whether this should be licensed as AGPL 3.0 or Apache 2.0. 

However either way I don't see a problem with carrying the license line for imported cherry-picks.
I would prefer not to have to start having more complex cherry-picking guidelines where the process is different for different repos.

tomponline:
> @MusicDin were these tested as i note: [lxc/lxc-ci#813](https://github.com/lxc/lxc-ci/issues/813)

@MusicDin ?

MusicDin:
Yep, I'm still testing the recipes. Will update the PR with required cherry picks once done. Moving to draft.

images/ubuntu.yaml
```diff
@@ -346,6 +346,7 @@ files:
   - kinetic
   - lunar
```
simondeziel, line 347: Time to drop kinetic (22.10) and lunar (23.04) as they are both EOL https://fridge.ubuntu.com/2023/12/15/ubuntu-23-04-lunar-lobster-reaches-end-of-life-on-january-25-2024/

Maybe there's a commit doing just that that needs to be cherry-picked?

simondeziel (approved):

simondeziel:
I'm not sure [images/ubuntu: Re-enable legacy cloud-init in Ubuntu](https://github.com/canonical/lxd-ci/pull/69/commits/839baf6f0024c89ab536711ec5689866d9848a85) is actually needed as I'd assume this was done because cloud-init expects a LXD data source instead of an Incus one?

Anyway, if we do need it we should mangle the changelog with `s/Incus/LXD/g`.

MusicDin:
There is no particular reason for those being split. I'll merge.

MusicDin:
Merged to #71

PR #74, opened 2024-02-21
tests: Check that we can launch a stateful VM without  `size.state`

gabrielmougard:
Requires https://github.com/canonical/lxd/pull/12912

* Here, we test that we can override the root `size.state` without stopping the stateful VM.
* Once the device is overridden, check that we can do stateful stop / start on the VM.


tests/storage-vm
```diff
@@ -484,6 +484,20 @@ for poolDriver in $poolDriverList; do
 
         lxc delete -f v1
 
+        echo "==> Checking start / stop operations with root size.state being unset"
+        # We should be able to start a VM with migration.stateful=true eventhough the root `size.state` config is unset
+        lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm --config migration.stateful=true -s "${poolName}"
+        # However, we should not be able to statefully stop the VM
+        ! lxc stop v1 --stateful
```
simondeziel, line 491:
```suggestion
        ! lxc stop v1 --stateful || false
```

simondeziel (changes requested):
LGTM with the tiny suggested change, thanks!

tests/storage-vm
```diff
@@ -484,6 +484,20 @@ for poolDriver in $poolDriverList; do
 
         lxc delete -f v1
 
+        echo "==> Checking start / stop operations with root size.state being unset"
+        # We should be able to start a VM with migration.stateful=true eventhough the root `size.state` config is unset
+        lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm --config migration.stateful=true -s "${poolName}"
```
tomponline, line 489: @simondeziel im going to merge this PR as I see this pattern occurring a lot.
But can we somehow avoid needing to specify the default value for TEST_IMG in every test and instead default to ubuntu-minimal-daily:22.04 in a single place where TEST_IMG is defined and just expect it to always be populated?

tomponline (dismissed):

tests/storage-vm
```diff
@@ -484,6 +484,20 @@ for poolDriver in $poolDriverList; do
...
+        # However, we should not be able to statefully stop the VM
+        ! lxc stop v1 --stateful || false
+        # But we should be able to change the root size.state config without stopping the VM
+        lxc config device override v1 root size.state=2GiB
+        # And then we should be able to statefully stop the VM
+        lxc stop v1 --stateful
+        # And then we should be able to statefully start the VM again
+        lxc start v1
```
tomponline, line 497: @gabrielmougard can you tweak this test to wait until the VM agent has restarted to confirm it was successfully restored.

simondeziel: Would it make sense to also compare the boot ID to make sure the VM was live-restored and not booted from 0?
`/proc/sys/kernel/random/boot_id` has a different UUID one each boot.

simondeziel: @gabrielmougard `waitInstanceBooted()` should let you wait for the VM to be usable in the `lxc exec` sense of usable.

tests/storage-vm
tomponline: Thanks!

tests/storage-vm
tomponline: Sounds good idea to me!

gabrielmougard: fixed

simondeziel (approved):
LGTM, thanks.

tomponline:
@gabrielmougard needs a rebase

tests/storage-vm
```diff
@@ -484,6 +484,29 @@ for poolDriver in $poolDriverList; do
 
         lxc delete -f v1
 
+        echo "==> Checking start / stop operations with root size.state being unset"
```
tomponline, line 487: @gabrielmougard this should check the same API extension as https://github.com/canonical/lxd-ci/pull/75 so it skips on 5.0/edge cc @simondeziel

simondeziel: I'll try to do a better job at catching those API checks, thanks for being so vigilant Tom.

simondeziel (approved):

tests/storage-vm
```diff
@@ -517,6 +517,31 @@ for poolDriver in $poolDriverList; do
         # Finally, unset the value to avoid any potential side effects on other tests.
         lxc config unset instances.migration.stateful
 
+        if hasNeededAPIExtension instances_migration_stateful; then
```
simondeziel, line 520: Looks like this needs to be quite a few lines before that. See the CI failures.

simondeziel (changes requested):

simondeziel (approved):

tomponline:
needs a rebase

tests/storage-vm

PR #75, opened 2024-02-21
tests: Check `instances.migration.stateful` effect on VMs' `migration.stateful` parameter

gabrielmougard:
Requires https://github.com/canonical/lxd/pull/12832

### DIFF
--- a/tests/storage-vm
+++ b/tests/storage-vm
@@ -484,6 +484,35 @@ for poolDriver in $poolDriverList; do
 
         lxc delete -f v1
 
+        echo "==> Checking setting instances.migration.stateful for VMs"
+        # Check that migration.stateful is set in the instance's expanded config to the value of `instances.migration.stateful`
+        # if no local instance config nor profile value of `migration.stateful` is set.
+
+        lxc config set instances.migration.stateful=true
+        lxc init "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm -s "${poolName}"
+
+        lxc query /1.0/instances/v1 | jq -r '.expanded_config["migration.stateful"]' | grep -xF true
+        lxc delete v1 -f
+        lxc config set instances.migration.stateful=false
+        lxc init "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm -s "${poolName}"
+        lxc query /1.0/instances/v1 | jq -r '.expanded_config["migration.stateful"]' | grep -xF false
+        lxc delete v1 -f
+
+        lxc config set instances.migration.stateful=true
+        lxc init "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm --config migration.stateful=false -s "${poolName}"
+        lxc query /1.0/instances/v1 | jq -r '.expanded_config["migration.stateful"]' | grep -xF false # the instance local config should take precedence
+        lxc delete v1 -f
+
+        lxc config set instances.migration.stateful=false
+        lxc profile copy default stateful_profile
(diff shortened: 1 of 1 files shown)


tests/storage-vm
```diff
@@ -484,6 +484,32 @@ for poolDriver in $poolDriverList; do
...
         lxc delete -f v1
 
+        echo "==> Checking setting instances.migration.stateful for VMs"
+        # Check that migration.stateful is set in the instance's expanded config to the value of `instances.migration.stateful`
+        # if no local instance config nor profile value of `migration.stateful` is set.
+
+        lxc config set instances.migration.stateful=true
+        lxc init "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm -s "${poolName}"
+
+        lxc query /1.0/instances/v1 | jq '.expanded_config["migration.stateful"]' | grep -q true
```
simondeziel, line 494:
```suggestion
        lxc query /1.0/instances/v1 | jq -r '.expanded_config["migration.stateful"]' | grep -xF true
```
I would prefer to use `jq` `-r` to avoid the quoting and then use a full line pattern matching with `grep`. No silencing `grep`'s output is also useful when debugging using log output.

tests/storage-vm
```diff
@@ -484,6 +484,32 @@ for poolDriver in $poolDriverList; do
...
+        lxc delete v1 -f
+        lxc config set instances.migration.stateful=false
```
simondeziel, line 496:
```suggestion
        lxc config unset instances.migration.stateful
```
This would ensure the default is indeed `false`.

tests/storage-vm
```diff
@@ -484,6 +484,32 @@ for poolDriver in $poolDriverList; do
...
+        lxc init "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm -s "${poolName}"
+        lxc query /1.0/instances/v1 | jq '.expanded_config["migration.stateful"]' | grep -q false
+        lxc delete v1 -f
+
+        lxc config set instances.migration.stateful=true
+        lxc init "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm --config migration.stateful=false -s "${poolName}"
+        lxc query /1.0/instances/v1 | jq '.expanded_config["migration.stateful"]' | grep -q false # the instance local config should take precedence
+        lxc delete v1 -f
+
+        lxc config set instances.migration.stateful=false
```
simondeziel, line 506: I'd probably `unset` here too.

gabrielmougard: But I think I forgot to do a `lxc config unset instances.migration.stateful` at the end to avoid potential side effects on the following tests.

tests/storage-vm

tests/storage-vm
simondeziel: It does, thanks for clarifying.

simondeziel (approved):

tomponline (approved):

tests/storage-vm
```diff
@@ -484,6 +484,35 @@ for poolDriver in $poolDriverList; do
 
         lxc delete -f v1
 
+        echo "==> Checking setting instances.migration.stateful for VMs"
```
tomponline, line 487: @gabrielmougard @simondeziel oh I just merged this, but then reaslised its not checking the API extension and skipping this test if server doesn't support it. Please can you update in a new PR or the 5.0/edge tests will fail. Thanks @gabrielmougard

gabrielmougard: @tomponline https://github.com/canonical/lxd-ci/pull/84

PR #79, opened 2024-02-21
Test container copy from external mounted storage

masnax:
Adds a test to catch snap-specific regressions in rsync, prompted by https://github.com/canonical/lxd/pull/12927

Should be merged after https://github.com/canonical/lxd/pull/12927

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -68,6 +68,7 @@ jobs:
         test:
           - cgroup
           - cluster
+          - container-copy
           - cpu-vm
           - devlxd-vm
           - docker
--- a/tests/container-copy
+++ b/tests/container-copy
@@ -1,0 +1,69 @@
+#!/bin/sh
+set -eu
+
+# Tests for snap-specific regressions when rsyncing a container whose storage pool is mounted in the snap confinement.
+
+# Install and setup LXD.
+install_lxd
+lxd init --auto
+
+# Create a loop device that we can mount as a source for a dir storage pool.
+test_dir="$(mktemp -d XXX.test)"
+loopimg="$(mktemp "${test_dir}/XXX.img")"
+truncate -s 5G "${loopimg}"
+loopdev="$(losetup --show --find "${loopimg}")"
+
+extra_cleanup() {
+  # Clean up LXD and remove the loop device.
+  lxc remote remove target
+  lxc delete target -f
+  losetup --detach "${loopdev}"
+  rm -rf "${test_dir}"
+}
+
+# Begin Test.
+set -x
+
+# Launch a VM which will be the target of the copy.
+lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" target --vm
+waitInstanceBooted target
+
+# Add the target VM as a remote LXD.
+lxc exec target -- snap install lxd --channel "${LXD_SNAP_CHANNEL}"
+lxc exec target -- lxd init --auto
+lxc exec target -- lxc config set core.https_address=:8443
+token="$(lxc exec target -- lxc config trust add --name host --quiet)"
+lxc remote add target "${token}" --accept-certificate
+
+# Run the test for various filesystems.
+for fs in "xfs" "btrfs" "ext4" ; do
+  mkfs.${fs} "${loopdev}"
+  disk_dir="$(mktemp -d "${test_dir}/XXX.disk")"
+  mount "${loopdev}" "${disk_dir}"
+  rm -rf "${disk_dir}/lost+found"
+
+  # Create a storage pool with the mounted disk. It will be mounted in the snap, but not on the hostfs.
+  lxc storage create "${fs}" dir source="${PWD}/${disk_dir}"
+
+  # Create a container which we will copy to another target.
+  lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" c1 --storage "${fs}"
+
+  # Copy the container to the remote LXD.
+  lxc copy c1 target:c1 --stateless --refresh --mode=relay
+
+  # Ensure the container exists.
+  lxc config show target:c1 > /dev/null
+
+  # Clean up the storage pool for this run.
+  lxc delete c1 -f
+  lxc delete target:c1 -f
+  lxc storage delete "${fs}"
+
+  # Reset the loop device.
+  umount "${loopdev}"
+  wipefs -a "${loopdev}"
+done
+
+
+# shellcheck disable=SC2034
+FAIL=0
--- a/tests/main-openstack
+++ b/tests/main-openstack
@@ -85,6 +85,9 @@ run_test jammy default tests/storage-disks-vm "${lxd_snap_channel}"
 run_test jammy default tests/storage-vm "${lxd_snap_channel}"
 run_test jammy default tests/storage-volumes-vm "${lxd_snap_channel}"
 
+# containers
+run_test jammy default tests/container-copy "${lxd_snap_channel}"
+
 # vm-nesting
 run_test jammy default tests/vm-nesting "${lxd_snap_channel}"
 run_test jammy hwe tests/vm-nesting "${lxd_snap_channel}"
--- a/tests/main-testflinger
+++ b/tests/main-testflinger
@@ -86,6 +86,9 @@ run_test jammy default tests/storage-disks-vm "${lxd_snap_channel}"
 run_test jammy default tests/storage-vm "${lxd_snap_channel}"
 run_test jammy default tests/storage-volumes-vm "${lxd_snap_channel}"
 
+# containers
+run_test jammy default tests/container-copy "${lxd_snap_channel}"
+
 # vm-nesting
 run_test jammy default tests/vm-nesting "${lxd_snap_channel}"
 run_test jammy hwe tests/vm-nesting "${lxd_snap_channel}"


tomponline (commented):
@simondeziel @masnax i cant see the test running on GH workflows - we should now see it passing on latest/edge but failing on 5.0/edge. Thanks

tests/container-copy
```diff
@@ -0,0 +1,64 @@
...
+exit 1
+"
+
+# Add the target VM as a remote LXD.
+target_addr="$(lxc ls -f csv -c 4 target | grep -v lxdbr | cut -d' ' -f1 | sed 's/"//g')"
+lxc exec target -- snap install lxd --channel "${LXD_SNAP_CHANNEL}"
+lxc exec target -- lxd init --auto
+lxc exec target -- lxc config set core.https_address=:8443
+lxc exec target -- lxc config set core.trust_password=lxd123
+lxc remote add target "https://${target_addr}:8443" --accept-certificate --password "lxd123"
```
simondeziel, line 48: I'd rather avoid adding dependency on password auth at this point. It is already waiting in the death row ;)

tests/container-copy
```diff
@@ -0,0 +1,64 @@
...
+set -x
+
+# Create a storage pool with the mounted disk.
+lxc storage create dir dir source="${PWD}/${disk_dir}"
+
+# Create a container which we will copy to another target.
+lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" c1 --storage dir
+
+# Launch a VM which will be the target of the copy.
+lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" target --vm
```
simondeziel, line 31: Does it have to be a VM?

tests/container-copy
```diff
@@ -0,0 +1,64 @@
...
+lxc exec target -- lxd init --auto
+lxc exec target -- lxc config set core.https_address=:8443
+lxc exec target -- lxc config set core.trust_password=lxd123
+lxc remote add target "https://${target_addr}:8443" --accept-certificate --password "lxd123"
+
+# Copy the container to the remote LXD.
+lxc copy c1 target:c1 --stateless --refresh --mode=relay
+
+# Clean up LXD.
+lxc rm c1 target -f
```
simondeziel, line 54: In tests, we try to use the main command verb (`s/rm/delete/`).

tests/container-copy
masnax: Trying this with a container causes a bunch of permission denied errors. I'm not sure if that's expected or not, as it happens with non-externally mounted storage as well:

```
Error: Error transferring instance data: Failed migration on target: Failed creating instance on target: Rsync receive failed: /var/snap/lxd/common/lxd/storage-pools/default/containers/c1/: [exit status 23] (rsync: [generator] mknod "/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/dev/console" failed: Operation not permitted (1)
rsync: [generator] mknod "/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/dev/full" failed: Operation not permitted (1)
rsync: [generator] mknod "/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/dev/null" failed: Operation not permitted (1)
rsync: [generator] mknod "/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/dev/ptmx" failed: Operation not permitted (1)
rsync: [generator] mknod "/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/dev/random" failed: Operation not permitted (1)
rsync: [generator] mknod "/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/dev/tty" failed: Operation not permitted (1)
rsync: [generator] mknod "/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/dev/urandom" failed: Operation not permitted (1)
rsync: [generator] mknod "/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/dev/zero" failed: Operation not permitted (1)
rsync: [receiver] rsync_xal_set: lsetxattr("/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/var/log/journal","trusted.SGI_ACL_DEFAULT") failed: Operation not permitted (1)
rsync: [receiver] rsync_xal_set: lsetxattr("/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/var/log/journal","trusted.SGI_ACL_FILE") failed: Operation not permitted (1)
rsync: [receiver] rsync_xal_set: lsetxattr("/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/var/log/journal/5c9941407f4e46c5b6c652644b1c31b2","trusted.SGI_ACL_DEFAULT") failed: Operation not permitted (1)
rsync: [receiver] rsync_xal_set: lsetxattr("/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/var/log/journal/5c9941407f4e46c5b6c652644b1c31b2","trusted.SGI_ACL_FILE") failed: Operation not permitted (1)
rsync: [receiver] rsync_xal_set: lsetxattr("/var/snap/lxd/common/lxd/storage-pools/default/containers/c1/rootfs/var/log/journal/5c9941407f4e46c5b6c652644b1c31b2/.system.journal.wEww52","trusted.SGI_ACL_FILE") failed: Operation not permitted (1)
```
I also tried with `security.nesting=true` but that didn't change anything either.

masnax: @tomponline Any thoughts about this one?

simondeziel: Those errors sound as if the FS was mounted with `nodev` and without XATTR support, weird.

masnax: I've removed the non-`mknod` errors as those were an issue with the build of LXD I was using for the test. On `latest/edge` the only errors are the `mknod` ones.

tests/container-copy
```diff
@@ -0,0 +1,64 @@
...
+cp /root/go/bin/lxd /var/snap/lxd/common/lxd.debug
+systemctl reload snap.lxd.daemon
+lxd waitready --timeout 100
+
+# Create a loop device that we can mount as a source for a dir storage pool.
+test_dir="$(mktemp -d XXX.test)"
+loopimg="$(mktemp "${test_dir}/XXX.img")"
+truncate -s 5G "${loopimg}"
+loopdev="$(losetup --show --find "${loopimg}")"
+mkfs.xfs "${loopdev}"
```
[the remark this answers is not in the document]
masnax: I've added a loop that checks `xfs` `ext4` and `btrfs`. 

For `ext4` we also have to delete `lost+found` otherwise LXD will complain that the source is non-empty.

tests/container-copy
```diff
@@ -0,0 +1,65 @@
...
+
+  # Copy the container to the remote LXD.
+  lxc copy c1 target:c1 --stateless --refresh --mode=relay
+
+  # Ensure the container exists.
+  lxc config show target:c1 > /dev/null
+
+  # Clean up the storage pool for this run.
+  lxc delete c1 -f
+  lxc exec target -- sh -c "lxc delete -f \$(lxc ls -f csv -c n)"
```
simondeziel, line 50: Would `lxc delete -f target:c1` work?

tests/container-copy
```diff
@@ -0,0 +1,65 @@
...
+
+# Run the test for various filesystems.
+for fs in "xfs" "btrfs" "ext4" ; do
+  mkfs.${fs} "${loopdev}"
+  disk_dir="$(mktemp -d "${test_dir}/XXX.disk")"
+  mount "${loopdev}" "${disk_dir}"
+  rm -rf "${disk_dir}/lost+found"
+
+  # Create a storage pool with the mounted disk.
+  lxc storage create dir dir source="${PWD}/${disk_dir}"
```
simondeziel, line 37: To ease debugging, it would help to embed the FS name in the storage pool name, please.

simondeziel (commented):
Thanks for adding the multiple FSes! LGTM except the password auth thingy.

tests/container-copy
masnax: Good point :)

tests/container-copy
masnax: I've updated this to use cert auth. In case the test was run without LXD having generated the config directory, I first have to call `lxc remote add <something>` to get the client to create certs.

simondeziel: Thanks, I should have provided you with a pointer to pylxd that does just that (https://github.com/canonical/pylxd/blob/main/integration/run-integration-tests#L18-L19) sorry.

simondeziel (approved):
LGTM and it's pretty fast to run so thanks!

tests/container-copy
tomponline: Unprivileged containers cannot mknod.
You may be able to workaround this using a privileged container, e.g using some or all of this.

```
  raw.lxc: |-
    lxc.cgroup2.devices.allow = b 259:* rw
    lxc.cgroup.devices.allow = b 259:* rw
  security.nesting: "true"
  security.privileged: "true"
  ```
  
  We use this for running lxd-imagebuilder inside a container.

Comment (tomponline) [reply tests/container-copy:18]: Or potentially using security.syscalls.intercept.mknod see https://documentation.ubuntu.com/lxd/en/latest/syscall-interception/#available-system-calls

File: tests/container-copy [hunk 7]
```diff
@@ -0,0 +1,70 @@
...
+waitInstanceBooted target
+
+# Add the target VM as a remote LXD.
+target_addr="$(lxc ls -f csv -c 4 target | grep -v lxdbr | cut -d' ' -f1 | sed 's/"//g')"
+lxc exec target -- snap install lxd --channel "${LXD_SNAP_CHANNEL}"
+lxc exec target -- lxd init --auto
+lxc exec target -- lxc config set core.https_address=:8443
+
+# Use `lxc remote add` to generate client certs, in case they don't exist.
+! lxc remote add generate-certs not-a-url --accept-certificate 2>/dev/null || true
```
tomponline, line 28: Whats happening here, i read the comment and still dont follow.

tests/container-copy
```diff
@@ -0,0 +1,70 @@
...
+  lxc delete c1 -f
+  lxc delete target:c1 -f
+  lxc storage delete "${fs}"
+
+  # Reset the loop device.
+  umount "${loopdev}"
+  wipefs -a "${loopdev}"
+done
+
+# Clean up LXD and remove the loop device.
```
tomponline, line 63: @simondeziel should this be a cleanup function?

tests/container-copy
```diff
@@ -0,0 +1,70 @@
+#!/bin/sh
+set -eu
+
```
tomponline, line 3: Could you add a comment here explaining what this test is for please. The *why*

tests/container-copy
tomponline: > I'd rather avoid adding dependency on password auth at this point. It is already waiting in the death row ;)

@simondeziel we'll need to keep tests for password auth at least for LXD 5.0 and LXD 5.21 btw to avoid regressions as its still supported in those series. Please can you ensure we are still testing it somewhere in the snap based tests. Thanks

tomponline: Maybe here would be a good place to do it.

.github/workflows/tests.yml
```diff
@@ -85,6 +85,7 @@ jobs:
           - "storage-vm lvm-thin"
           - "storage-vm zfs"
           - storage-volumes-vm
+          - container-copy
```
simondeziel, line 88: Since you are going to alter this PR at least one more time to address Tom's feedback, would you mind alpha-sorting that list? </nitpick>

tests/container-copy
simondeziel: That `security.syscalls.intercept.mknod` config sounds nice indeed however, my initial question of "why a VM" was because I feared the test would be slow. Turns out it's quite fast (4-5 minutes) so my concern was ill-founded.

Now if that `mknod` config works, it would be even faster as it would save downloading the 2nd image (for the VM) as it could use the same image twice.

tests/container-copy
simondeziel: @tomponline tests defining an `extra_cleanup()` function will have their cleanup done only on successful runs which is fine but they will run that cleanup with `set +e` meaning if a cleanup step fails, it won't fail the test.

I'm now questioning my own decision of doing this as being able to cleanup properly feels important too.

tests/container-copy
simondeziel, line 28:
```suggestion
# Try to add a bogus remote to force lxc into generating a client crt/key pair if none was found
if ! [ -e ~/snap/lxd/common/config/client.crt ]; then
  lxc remote add generate-certs not-a-url --accept-certificate 2>/dev/null || true
fi
```
Would that be good enough?

tomponline: > I'm now questioning my own decision of doing this as being able to cleanup properly feels important too.

Yeah cleanup doesnt normally take long either.

tomponline: >   lxc remote add generate-certs not-a-url --accept-certificate 2>/dev/null || true
I'm thinking this is going to fail if/when we fix https://github.com/canonical/lxd/issues/12414

I'm not quite following why we need this "fudge"?

tomponline: > Turns out it's quite fast (4-5 minutes) so my concern was ill-founded.

I'm happy with a VM for now too.

simondeziel: > > lxc remote add generate-certs not-a-url --accept-certificate 2>/dev/null || true
> > I'm thinking this is going to fail if/when we fix [canonical/lxd#12414](https://github.com/canonical/lxd/issues/12414)
> 
> I'm not quite following why we need this "fudge"?

It's because the client cert needs to be added to the `target`'s trust for it to become a usable remote.
```
lxc file push ~/snap/lxd/common/config/client.crt target/root/
lxc exec target -- lxc config trust add test /root/client.crt
```

Having said that, it feels like a good case for a [trust token](https://documentation.ubuntu.com/lxd/en/latest/howto/server_expose/#authenticate-with-the-lxd-server).

masnax: > > lxc remote add generate-certs not-a-url --accept-certificate 2>/dev/null || true
> > I'm thinking this is going to fail if/when we fix [canonical/lxd#12414](https://github.com/canonical/lxd/issues/12414)
> 
> I'm not quite following why we need this "fudge"?

The client uses `~/snap/lxd/common/config` or `~/.config/lxc` to store client certs. To be able to use a remote there has to be a keypair present here. It doesn't exist by default when you `snap install lxd`. When we use `lxc remote add` we automatically create that directory and populate it before actually performing the action of adding the remote. 

In this case we are using cert auth so the client cert needs to have been trusted by the other system before we call `lxc remote add`, so we have to use it once to generate the certs, then move the certs over and trust them on the other side, and use it again with the actual url to add the remote.

simondeziel: @masnax a trust token would be simpler and avoid moving files around.

simondeziel (approved):

PR #83, opened 2024-02-26
tests/storage-vm: Test that we avoid deleting existing target directories not created by LXD when unmounting disks

gabrielmougard:
Related to https://github.com/canonical/lxd/pull/12700

### DIFF
--- a/tests/storage-vm
+++ b/tests/storage-vm
@@ -242,6 +242,39 @@ for poolDriver in $poolDriverList; do
         echo "==> Deleting VM"
         lxc delete -f v1
 
+        if hasNeededAPIExtension disk_state_created; then
+                # Check that we don't remove an existing target directory when unmounting a disk device if the original dir hasn't been created by LXD
+                echo "==> Avoid deleting existing target directories not created by LXD when unmounting disks."
+                mkdir -p "/tmp/lxd-test-${poolName}/empty-dir"
+                lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm -s "${poolName}"
+                waitInstanceReady v1
+
+                lxc exec v1 -- test -d /opt # The directory already exists and is not created by LXD.
+                lxc config device add v1 empty-dir disk source="/tmp/lxd-test-${poolName}/empty-dir" path=/opt
+                # Check that no `volatile.<deviceName>.last_state.created` has been set.
+                if [ -n "$(lxc config get v1 volatile.empty-dir.last_state.created)"]; then
+                        echo "==> volatile.<deviceName>.last_state.created should not be set"
+                        false
+                fi
+
+                lxc config device remove v1 empty-dir
+                lxc exec v1 -- test -d /opt # /opt is not removed as it was not created by LXD.
+
+                lxc config device add v1 empty-dir disk source="/tmp/lxd-test-${poolName}/empty-dir" path=/tmp/new-dir
+                # Check that the volatile.<deviceName>.last_state.created is set.
+                targetPath=$(lxc config get v1 volatile.empty-dir.last_state.created)
+                if [ "${targetPath}" != "/tmp/new-dir" ]; then
+                        echo "==> volatile.<deviceName>.last_state.created is not set to the correct value"
+                        false
+                fi
+
+                lxc config device remove v1 empty-dir
+                ! lxc exec v1 -- test -d /tmp/new-dir || false # /tmp/new-dir is removed as it was created by LXD.
+
+                lxc delete -f v1
+                rmdir "/tmp/lxd-test-${poolName}/empty-dir"
+        fi
+
         # Create directory with a file for directory disk tests.
         mkdir "/tmp/lxd-test-${poolName}"
 


tests/storage-vm
```diff
@@ -242,6 +242,49 @@ for poolDriver in $poolDriverList; do
...
+                ! lxc exec v1 -- mkdir /opt || false # The directory already exists and is not created by LXD.
+                lxc config device add v1 empty-dir disk source="/tmp/lxd-test-${poolName}/empty-dir" path=/opt
+                # Check that no `volatile.<deviceName>.last_state.created` has been set.
+                if [ -n "$(lxc config get v1 volatile.empty-dir.last_state.created)"]; then
+                        echo "==> volatile.<deviceName>.last_state.created should not be set"
+                        false
+                fi
+
+                lxc config device remove v1 empty-dir
+                lxc exec v1 -- stat /opt # /opt is not removed as it was not created by LXD.
```
simondeziel, line 261:
```suggestion
                lxc exec v1 -- test -d /opt # /opt is not removed as it was not created by LXD.
```

tests/storage-vm
```diff
@@ -242,6 +242,49 @@ for poolDriver in $poolDriverList; do
...
+
+                lxc exec v1 -- mkdir -p /tmp/empty-dir # Create a new directory to check that it is not removed when unmounting the disk.
```
simondeziel, line 263:
```suggestion
                lxc exec v1 -- mkdir /tmp/empty-dir # Create a new directory to check that it is not removed when unmounting the disk.
```

`/tmp` has to be there, otherwise we have other problems.

tests/storage-vm
```diff
@@ -242,6 +242,49 @@ for poolDriver in $poolDriverList; do
...
+                lxc config device add v1 empty-dir disk source="/tmp/lxd-test-${poolName}/empty-dir" path=/tmp/empty-dir
+                # Check that no `volatile.<deviceName>.last_state.created` has been set.
+                if [ -n "$(lxc config get v1 volatile.empty-dir.last_state.created)"]; then
+                        echo "==> volatile.<deviceName>.last_state.created should not be set"
+                        false
+                fi
+
+                lxc config device remove v1 empty-dir
+                lxc exec v1 -- stat /tmp/empty-dir
```
simondeziel, line 272:
```suggestion
                lxc exec v1 -- test -d /tmp/empty-dir
```

tests/storage-vm
```diff
@@ -242,6 +242,49 @@ for poolDriver in $poolDriverList; do
...
         lxc delete -f v1
 
+        if hasNeededAPIExtension disk_state_created; then
+                # Check that we don't remove an existing target directory when unmounting a disk device if the original dir hasn't been created by LXD
+                echo "==> Avoid deleting existing target directories not created by LXD when unmounting disks."
+                mkdir -p "/tmp/lxd-test-${poolName}/empty-dir"
+                lxc launch "${TEST_IMG:-ubuntu-minimal-daily:22.04}" v1 --vm -s "${poolName}"
+                waitInstanceReady v1
+
+                ! lxc exec v1 -- mkdir /opt || false # The directory already exists and is not created by LXD.
```
simondeziel, line 252:
```suggestion
                lxc exec v1 -- test -d /opt # The directory already exists and is not created by LXD.
```

tests/storage-vm
```diff
@@ -242,6 +242,49 @@ for poolDriver in $poolDriverList; do
...
+                lxc config device add v1 empty-dir disk source="/tmp/lxd-test-${poolName}/empty-dir" path=/tmp/new-dir
+                # Check that the volatile.<deviceName>.last_state.created is set.
+                targetPath=$(lxc config get v1 volatile.empty-dir.last_state.created)
+                if [ "${targetPath}" != "/tmp/new-dir" ]; then
+                        echo "==> volatile.<deviceName>.last_state.created is not set to the correct value"
+                        false
+                fi
+
+                lxc config device remove v1 empty-dir
+                ! lxc exec v1 -- stat /tmp/new-dir || false # /tmp/new-dir is removed as it was created by LXD.
```
simondeziel, line 283:
```suggestion
                ! lxc exec v1 -- test -d /tmp/new-dir || false # /tmp/new-dir is removed as it was created by LXD.
```

simondeziel (changes requested):
I don't see the difference between the test with `/opt` and `/tmp/empty-dir` as both are **not** created by LXD.

tests/storage-vm
gabrielmougard: fixed

tests/storage-vm
gabrielmougard: fixed

tests/storage-vm
gabrielmougard: fixed

tests/storage-vm
gabrielmougard: fixed

tests/storage-vm

gabrielmougard:
> I don't see the difference between the test with `/opt` and `/tmp/empty-dir` as both are **not** created by LXD.

you're right, these two cases test the same thing, removing the second one.

simondeziel (approved):

simondeziel:
@gabrielmougard @tomponline I wonder how much of that whole test (not just the bits added by Gabriel) also apply to container. Should we consider creating a `storage-ctn` variant of the whole script?

simondeziel:
We should wait for https://github.com/canonical/lxd/pull/12700 to be merged so that we can trigger a run in here before merging.

tomponline:
Not needed atm

PR #86, opened 2024-02-27
Improvements to `tests/cgroup`

simondeziel:


simondeziel:
@tomponline this was tested on PS6 against a Focal machine:

```
$ ./bin/openstack-run jammy default tests/cgroup latest/edge
```

So it seems there is no regression from Go 1.22 when used on an old glibc host, presumably thanks to `core22` shipping with a new glibc.

tomponline (approved):

tomponline:
> @tomponline this was tested on PS6 against a Focal machine:
> 
> ```
> $ ./bin/openstack-run jammy default tests/cgroup latest/edge
> ```
> 
> So it seems there is no regression from Go 1.22 when used on an old glibc host, presumably thanks to `core22` shipping with a new glibc.

Excellent thanks!

PR #88, opened 2024-02-28
Fix image builds

MusicDin:


bin/test-image
```diff
@@ -75,11 +75,6 @@ if [ "${TYPE}" = "vm" ]; then
 
     INSTANCES="${TEST_IMAGE}"
 
-    # Handle distros needing the agent drive.
```
tomponline, line 78: Please can you update your commit message so that it has a *why* and not just a *what*? Thanks

MusicDin: I've added description to the commit:

RHEL falvored ditros do not ship 9p kernel driver, but LXD mounts the agent over virtiofs.

tomponline: Thats not strictly true, we export config drive by 9p and virtiofs.

tomponline (approved):

bin/test-image
MusicDin: Yes. What I meant is that even though the RHEL flavor distros do not support 9p, LXD is still able to mount the agent over virtiofs.

PR #89, opened 2024-02-28
Configurable snap tracks

simondeziel:
I did a bit of cleanup and enabled `tests/interception` on `5.0/*` as it works there and I couldn't remember why I decided to exclude it.


tomponline:
Thanks!

PR #92, opened 2024-03-04
Wait instance to be ready

MusicDin:
Wait for instance to be ready when testing image builds instead of waiting for a fixed time.
We will see whether this approach will cause any significant issues.

Also this PR disables Debian Sid builds (for now).

### DIFF
--- a/bin/test-image
+++ b/bin/test-image
@@ -1,5 +1,8 @@
-#!/bin/sh
+#!/bin/bash
 set -eu
+
+# Source helper functions.
+. $(realpath $(dirname "$0")/helpers)
 
 # Check input arguments.
 if [ "${1:-}" = "" ] || [ "${2:-}" = "" ] || [ "${3:-}" = "" ] || [ "${4:-}" = "" ] || [ "${5:-}" = "" ]; then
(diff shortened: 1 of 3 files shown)


bin/test-image
```diff
@@ -145,10 +165,13 @@ for i in ${INSTANCES}; do
     lxc start "${i}"
 done
 
-# Wait for things to settle.
 echo "==> Waiting for instances to start"
-[ "${TYPE}" = "container" ] && sleep 1m
-[ "${TYPE}" = "vm" ] && sleep 3m
+for i in ${INSTANCES}; do
+    waitInstanceReady "${i}"
+done
```
simondeziel, line 171: We could tweak `bin/helpers` `waitInstanceReady()`  to consult a `MIN_PROC_COUNT` variable and then use something like this:
```suggestion
for i in ${INSTANCES}; do
    MIN_PROC_COUNT=1 waitInstanceReady "${i}"
done
```

simondeziel (commented):
I'd rather have the `waitInstanceReady` function from `bin/helpers` tweaked instead of duplicating it.

bin/test-image
simondeziel: In the above suggestion, I didn't mention but `bin/helpers` would need to look something like that:

```
# waitInstanceReady: waits for the instance to be ready (processes count > MIN_PROC_COUNT).
waitInstanceReady() (
  set +x
  maxWait=150
  instName="${1}"
  instProj="${2:-}"
  if [ -z "${instProj}" ]; then
    # Find the currently selected project.
    instProj="$(lxc project list -f csv | sed -n 's/^\([^(]\+\) (current),.*/\1/ p')"
  fi

  # Wait for the instance to report more than one process.
  processes=0
  for _ in $(seq "${maxWait}"); do
      processes="$(lxc info --project "${instProj}" "${instName}" | awk '{if ($1 == "Processes:") print $2}')"
      if [ "${processes}" -ge "${MIN_PROC_COUNT:-2}" ]; then
          return 0 # Success.
      fi
      sleep 1
  done

  echo "Instance ${instName} (${instProj}) not ready after ${maxWait}s"
  return 1 # Failed.
)
```

MusicDin: Yep, I've said that in description, but planned to first test this to see if it works (for all distros). 

Will fix anyway.

MusicDin: Thanks, very neat approach. 👍

tomponline (approved):

PR #93, opened 2024-03-05
Run image builds on self-hosted runners

MusicDin:
This should run image builds on the self-hosted runners. 

I've temporary disabled the majority of image builds to not waste the resources in case something goes wrong.

I'm not sure if `focal` runners are available, as those could be required for certain images (such as Orcale and AmazonLinux) if jammy ones do not support cgroups v1.

### DIFF
--- a/.github/workflows/image-almalinux.yml
+++ b/.github/workflows/image-almalinux.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   almalinux:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux, X64, jammy, large]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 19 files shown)


simondeziel (approved):
Looking at https://github.com/canonical/lxd-ci/actions/runs/8157515880/usage, it seems those build are quite fast. I don't know how they compare to those on GitHub provided runners?

According to [this](https://docs.google.com/document/d/1dwokjUyb2av9VibwEMURLYtulRrRaXKt116gYZXaFYg/edit?pli=1), those `large` Canonical runners have ~half of what GitHub runners provide. It's good if we can use them as there are 57 of those (https://git.launchpad.net/canonical-terraform-plans/tree/github-runners/environments/prod/main.tf#n87 and https://git.launchpad.net/canonical-terraform-plans/tree/github-runners/environments/prod-ps6/main.tf#n240).

tomponline (approved):

PR #95, opened 2024-03-06
Remove VM cpu limit in image tests

MusicDin:
Image build tests are currently failing because `large` self-hosted runners have only 2 CPUs and we cannot allocate more CPUs than available to the VM.

I would prefer not use larger runners (`xlarge`), since there are only 10 available.

### DIFF
--- a/bin/test-image
+++ b/bin/test-image
@@ -72,9 +72,9 @@ INSTANCES=""
 if [ "${TYPE}" = "vm" ]; then
     lxc init "${TEST_IMAGE}" "${TEST_IMAGE}" \
         --vm \
-        -c limits.cpu=4 \
         -c limits.memory=4GB \
         -c security.secureboot=false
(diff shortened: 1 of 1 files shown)


tomponline (approved):

simondeziel:
It's a bit late but I would have preferred to keep the CPU override. Here's some possible way to dynamically adapt to it:
```
NPROC="$(nproc)"
CPU_LIMIT="$((NPROC > 4 ? 4 : 1))"

lxc init --vm ... \
  -c limits.cpu="${CPU_LIMIT}"
```

This way, if the host has less than 5 cores, we only give 1 to the VM. If the hosts has 5+ cores, we give it 4 cores.

MusicDin:
Thanks for suggestion. However, we've reverted this change and as per our discussion it's probably better to fail if there is not enough cores then to debug why image tests are timing out.

PR #96, opened 2024-03-06
Increase wait instance ready timeout for image tests

MusicDin:
Currently, image VM tests are partially failing because instance is not ready in time (2 minutes). This increases wait instance timeout to 4 minutes in image tests.

Still, this may be a bit concerning as Alpine VMs are not ready within 2 minutes. I'm wondering how long will it take for *heavier* images.


MusicDin:
I have the same feeling, just wanted to prove it.

I'll revert self-hosted runners changes.

PR #98, opened 2024-03-06
tests/cpu-vm: reject `limits.kernel.*` for VMs

MggMuggins:
...and allow them in profiles. I chose not to go with @simondeziel's suggestion to use `--empty` as the tests fit well enough in `cpu-vm`, but I'm happy to spin up a new test if needed.

Resolves https://github.com/canonical/lxd/pull/13051

### DIFF
--- a/tests/cpu-vm
+++ b/tests/cpu-vm
@@ -19,6 +19,19 @@ poolDriver=dir
 echo "==> Create storage pool using driver ${poolDriver}"
 lxc storage create "${poolName}" "${poolDriver}"
 
+# limits.kernel.* aren't valid for VMs, but profiles with the key set should
+# still work
+lxc profile set default limits.kernel.nofile 50
+
+! lxc init v0 --vm --empty -c limits.kernel.cpu=46 -s "${poolName}" || false
+
+lxc init v0 --vm --empty -s "${poolName}"
+
+# limits.kernel.* only applies to containers (shouldn't work)
+! lxc config set v0 limits.kernel.as=1GiB || false
+
+lxc delete v0
+
 echo "==> Create ephemeral VM and boot"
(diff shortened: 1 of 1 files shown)


simondeziel:
@MggMuggins reusing an existing VM is indeed better than creating a new one, even empty. That said, the test currently passes for the wrong reason:

```
+ lxc config set v1 limits.kernel.as=1GiB
Error: Key "limits.kernel.as" cannot be updated when VM is running
```

You could possibly rework the test to replace `lxc launch` by `lxc init`, test `! lxc config set v1 limits.kernel.as... || false`, then `lxc start`.

Or go with a new empty VM that's only ever init'ed, config set failed then deleted.

MggMuggins:
:facepalm: Had some difficulties getting the tests to run at all and totally missed checking. Fixing...

simondeziel (approved):
LGTM, thanks! This currently fails because the corresponding lxd PR isn't merged yet but that's fine.

@tomponline that will be backported to 5.0 so no need for version detection/API extension detection logic, right?

tomponline:
> @tomponline that will be backported to 5.0 so no need for version detection/API extension detection logic, right?

Yeah this seems like something we can backport easily enough.

tomponline (changes requested):
Please can you rebase your PR so that we only see the final logical commit(s) rather than the evolution of the PR. We don't use squash merges, so we see all PR commit history in the main branch and want it to tell a logical story of changes.

MggMuggins:
Squashed & passed locally. Good to go so long as the CI agrees.

simondeziel:
Tests look good: working on `latest/edge` but failing where they should with `5.0/edge` that didn't get the backport yet.

tomponline (approved):
Thanks!

PR #101, opened 2024-03-08
VM: Add migration and refresh checks

roosterfish:
This PR adds checks to the `storage-vm` test suite for migration between the same and different storage drivers as well as local and remote refresh.

This is a follow up of https://github.com/canonical/lxd/pull/12743.


roosterfish:
Interesting, both Btrfs `5.0/edge` and `latest/edge` are blocking on `lxc copy v1 v2 --refresh`.

tomponline:
> Interesting, both Btrfs `5.0/edge` and `latest/edge` are blocking on `lxc copy v1 v2 --refresh`.

Are you able to take a look?

roosterfish:
> > Interesting, both Btrfs `5.0/edge` and `latest/edge` are blocking on `lxc copy v1 v2 --refresh`.
> 
> Are you able to take a look?

I am just running those tests for the Ceph RBD driver and found that only when migrating from a snapshot, the two volumes (block and fs) have the same size on the target:

```bash
root@thinkpad:/home/julian/dev/lxd-ci# lxc cp v1/snap0 localhost:v2
root@thinkpad:/home/julian/dev/lxd-ci# lxc shell m1
root@m1:~# rbd ls c -l
NAME                                                                                        SIZE     PARENT                                                                                        FMT  PROT  LOCK
...  
virtual-machine_v1                                                                          100 MiB  c/image_98f9a02ad850b4fe9b7d51b8b88c62556a4962d35afb7eacd31c666239421748_ext4@readonly          2            
virtual-machine_v1@snapshot_snap0                                                           100 MiB  c/image_98f9a02ad850b4fe9b7d51b8b88c62556a4962d35afb7eacd31c666239421748_ext4@readonly          2  yes       
virtual-machine_v1.block                                                                     10 GiB  c/image_98f9a02ad850b4fe9b7d51b8b88c62556a4962d35afb7eacd31c666239421748_ext4.block@readonly    2            
virtual-machine_v1.block@snapshot_snap0                                                      10 GiB  c/image_98f9a02ad850b4fe9b7d51b8b88c62556a4962d35afb7eacd31c666239421748_ext4.block@readonly    2  yes       
virtual-machine_v2                                                                          100 MiB                                                                                                  2            
virtual-machine_v2.block                                                                    100 MiB                                                                                                  2            

```

I am trying to find the root cause and take a look on the Btrfs one afterwards.

roosterfish:
Regarding the Ceph issue: 
I can confirm this was already a bug before we have changed the Ceph RBD refresh behavior. 

The `.block` volume gets created using the correct size on the target. But the source side starts sending the FS vol which will cause the target volume to be reduced in size when the diff gets applied on top. 

Working on a fix now.

roosterfish:
The `storage-vm` suite is running fine over the changes from this PR for Ceph RBD with the fixes from https://github.com/canonical/lxd/pull/13079.

PR #108, opened 2024-03-22
Publish images after successful build

MusicDin:
This PR enables image publishing.

### DIFF
--- a/.github/workflows/image-almalinux.yml
+++ b/.github/workflows/image-almalinux.yml
@@ -2,11 +2,16 @@ name: Build AlmaLinux Images
 
 on:
   workflow_dispatch:
-  workflow_call:
     inputs:
       publish:
         type: boolean
         default: false
+        description: Publish built image
+  workflow_call:
+    inputs:
+      publish:
+        type: boolean
+        default: true
         description: Publish built image
 
 jobs:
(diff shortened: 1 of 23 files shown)


simondeziel (approved):

.github/workflows/image-almalinux.yml
```diff
@@ -2,12 +2,17 @@ name: Build AlmaLinux Images
 
 on:
   workflow_dispatch:
-  workflow_call:
     inputs:
       publish:
         type: boolean
         default: false
         description: Publish built image
+  workflow_call:
```
tomponline, line 10: A comment explaining why would be useful dont you think?

tomponline (approved):

.github/workflows/image-almalinux.yml
MusicDin: We have a *master* workflow `.github/workflows/images.yml` that triggers other `image-*.yaml` workflows. This is why directive `workflows_call` is required.

Additionally, each `image-*.yaml` has a `workflow_dispatch` to enable us to trigger the build only for that specific image. For example, if we are fixing/updating Alpine builds, that we can trigger only Alpine workflow.

PR #110, opened 2024-03-22
Fix imageserver public key

MusicDin:


simondeziel (approved):
This was double checked to be the known host entry of the images container.

The previous known host was from the host VM :/

tomponline (approved):

PR #113, opened 2024-03-25
Use SFTP instead of SCP to upload images

MusicDin:
SCP flag `-s` (*Use the SFTP protocol for transfers rather than the original scp protocol.*) is not supported on older runners (specifically 20.04), therefore, use SFTP client for file transfer.


tomponline (approved):

MusicDin:
@tomponline tested on my end and it works as expected (both for 20.04 and 22.04 runners)

PR #115, opened 2024-03-26
Calculate files checksums after image build

MusicDin:
Calculate files checksums after image build and store them into SHA256SUMS file.

Example output:
```
74fe35937bcbce690c8d7b3b10fd6287009e2496c9bfd36bb872f007e68a4c80  disk.qcow2
a23f772bde56a57dd8aa4bd29940b30dc9460950749e14160e42c9e9deffcff6  image.yaml
0c66bc04acb33b267353dad46fc8f6cb21e8643a3f62a96e46eab63b9d27e6c8  lxd.tar.xz
1687249d5266d81a3839b0e4e45d28dd2fea846ee3d89bbd2a0bb312218f0d0b  rootfs.squashfs
07f917acb3d8b0b02cf3af2d909fa41d29ede324977cfc5d462df70fc6adee71  serial
```


bin/build-distro
```diff
@@ -53,3 +53,8 @@ fi
 
 sudo rm -Rf rootfs
 echo "${SERIAL}" > serial
+
+# Calculate checksums for all files within a directory.
+for file in *; do
+    sha256sum "${file}" >> SHA265SUMS
+done
```
simondeziel, line 60: The only reason I see to use a `for` loop is if you want to avoid any existing `SHA256SUMS` file i the current directory. If that's your goal, I'd add this:
```
for file in *; do
  [ "${file}" = "SHA256SUMS" ] && continue
  ...
done
```

Otherwise, a simple `sha256sum * > SHA265SUMS` would do.

MusicDin: Nice, didn't know that will work. Thanks

simondeziel (approved):

tomponline (approved):

PR #117, opened 2024-03-29
Upload image to temporary dir and then move it

MusicDin:
Upload contents of the built image to the temporary dir and only once fully uploaded move the directory to the final destination to avoid making partially uploaded images available in the simplestream index.

Closes https://github.com/canonical/lxd/issues/13191

### DIFF
--- a/.github/actions/image-upload/action.yml
+++ b/.github/actions/image-upload/action.yml
@@ -46,19 +46,14 @@ runs:
         VERSION=$(cat "${SRC_DIR}/serial")
 
         # Create directory structure that will be mirrored on the target server.
-        PRODUCT_PATH="${SRC_DIR}-upload/${IMG_DIR}"
-        mkdir -p "${PRODUCT_PATH}/${VERSION}"
+        mkdir -p "${SRC_DIR}-upload/${IMG_DIR}/.${VERSION}"
+        mv ${SRC_DIR}/* "${SRC_DIR}-upload/${IMG_DIR}/.${VERSION}"
 
-        # Move config.yaml file if exists.
-        if [ -f "${SRC_DIR}/config.yaml" ]; then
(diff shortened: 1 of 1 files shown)


.github/actions/image-upload/action.yml
```diff
@@ -45,20 +45,19 @@ runs:
...
+        DST_DIR="${IMG_DIR}/${VERSION}"
 
-        # Move image content.
-        mv ${SRC_DIR}/* "${PRODUCT_PATH}/${VERSION}"
+        # Create directory structure that will be mirrored on the target server.
+        mkdir -p "${SRC_DIR}-upload/${TMP_DIR}"
+        mv ${SRC_DIR}/* "${SRC_DIR}-upload/${TMP_DIR}"
 
         # Use SFTP to upload images to the server.
         sftp -P ${SSH_PORT} "${SSH_USER}@${SSH_HOST}" <<EOF
```
simondeziel, line 59: I'd use `-b -` to hint `sftp` it's in batch mode and should abort on any error.

.github/actions/image-upload/action.yml
```diff
@@ -45,20 +45,19 @@ runs:
...
             put -r ${SRC_DIR}-upload/*
```
simondeziel, line 60: `man sftp` doesn't mention `put -r`, only `put -R` so while both apparently work, I'd switch to the officially documented lower case one. I know `scp` uses `-r`, confusing :/

simondeziel (commented):
Great, thanks for making those upload batches atomic!

.github/actions/image-upload/action.yml
simondeziel: Yep, `put -R` seems to be what's in the man page.

simondeziel (approved):

simondeziel (approved):
LGTM, thanks!

tomponline (approved):

PR #118, opened 2024-04-02
images: Add simplestream configuration

MusicDin:
Add simplestream configuration for image requirements and release aliases.

### DIFF
--- a/images/ubuntu.yaml
+++ b/images/ubuntu.yaml
@@ -1,6 +1,25 @@
 image:
   distribution: "ubuntu"
 
+simplestream:
+  release_aliases:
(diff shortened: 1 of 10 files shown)


images/ubuntu.yaml
```diff
@@ -1,6 +1,25 @@
 image:
   distribution: "ubuntu"
 
+simplestream:
+  release_aliases:
+    xenial: 16.04
+    bionic: 18.04
+    focal: 20.04
+    jammy: 22.04
+    lunar: 23.04
```
simondeziel, line 10: 23.04 is EOL

PR #121, opened 2024-04-02
Add slackware image build

MusicDin:
This is only for the `current` version of Slackware.
Version `15.0` does not build, so there will be a follow up when I manage to get it working.


simondeziel (approved):
Thanks

PR #116, opened 2024-03-28
Tests: Add metrics certificate checks to the cluster suite

roosterfish:
Related to https://github.com/canonical/lxd/pull/13214.

This PR adds checks to validate if client and metrics certificates won't loose it's permission as defined before the cluster upgrade.

PR #125, opened 2024-04-03
Cluster upgrade cert tests

tomponline:
From https://github.com/canonical/lxd-ci/pull/116

### DIFF
--- a/bin/helpers
+++ b/bin/helpers
@@ -165,6 +165,14 @@ createPowerFlexPool() (
     powerflex.user.password="${POWERFLEX_PASSWORD}"
 )
 
+# createCertificateAndKey: creates a new key pair.
+createCertificateAndKey() (
+  key_file="${1}"
+  crt_file="${2}"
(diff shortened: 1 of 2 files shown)


markylaing (approved):
LGTM

MusicDin (approved):

tomponline:
@markylaing we expect the 5.21/edge cluster tests to fail because its testing upgrading from lxd 5.21/stable to 5.21/edge and so the initial setup of certs will be invalid, right?

markylaing:
> @markylaing we expect the 5.21/edge cluster tests to fail because its testing upgrading from lxd 5.21/stable to 5.21/edge and so the initial setup of certs will be invalid, right?

I think it will fail when checking the restricted status of the unrestricted metrics cert in 5.21 stable, because they are all considered restricted in that release. I don't think it'll fail to create the unrestricted cert though, just that the restricted status in the DB has no effect.

PR #131, opened 2024-04-08
tests/interception: add test for finit_module interception

mihalicyn:
https://github.com/canonical/lxd/pull/13151

### DIFF
--- a/tests/interception
+++ b/tests/interception
@@ -80,6 +80,58 @@ else
     echo "Skipping mount fuse tests as the container_syscall_intercept_mount_fuse API extension is missing"
 fi
 
+if hasNeededAPIExtension container_syscall_intercept_finit_module; then
+    # it can be any module which is present in the host filesystem and at the same time not used and can be unloaded
+    # good examples: slip, ip_nat_pptp, ip_gre, ip_conntrack_pptp (rarely used modules at the same time widely available)
+    MODULE_TO_TRY="nf_nat_ftp"
+    echo "Trying with module ... ${MODULE_TO_TRY}"
+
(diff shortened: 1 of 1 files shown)


simondeziel (approved):
LGTM but I didn't know the actual module file had to be present in the container trying to modprobe it. I guess it makes sense though.

In a real world scenario, the user would be expected to `apt-get install "linux-modules-$(uname -r)" "linux-modules-extra-$(uname -r)"`?

mihalicyn:
> LGTM but I didn't know the actual module file had to be present in the container trying to modprobe it. I guess it makes sense though.

it has to be present, because we use this file from inside the container to get module name from it. Then we use this module name to load the module on the host using file from the host. The `finit_module` syscall operates on the file descriptor, not on the module name. In any case, the input from the user for us is the module file.

>In a real world scenario, the user would be expected to apt-get install "linux-modules-$(uname -r)" "linux-modules-extra-$(uname -r)" in the instance?

Absolutely.

tomponline:
> LGTM but I didn't know the actual module file had to be present in the container trying to modprobe it. I guess it makes sense though.

The syscall doesn't include the module name - only the module file handle.

tomponline (approved):
This is great thanks.

Now we have this kernel module loading test, could you add something similar but tot test the non-ondemand mode is working (i.e just `linux.kernel_modules` defined)?

mihalicyn:
> Now we have this kernel module loading test, could you add something similar but tot test the non-ondemand mode is working (i.e just `linux.kernel_modules` defined)?

yes, it's a good task to improve test coverage

PR #138, opened 2024-04-11
Tests: Add check for refreshing additional snapshots

roosterfish:
With this change we actually test if a delta of snapshots is transferred accordingly for both local and remote VM copies.

Before the refresh was only testing if the volume itself can be refreshed as the number of snapshots on both source and target volume was identical due to the copy operation.

Part of https://github.com/canonical/lxd/pull/13305.


roosterfish:
The `storage-vm` tests on zfs (e.g. https://github.com/canonical/lxd-ci/actions/runs/8644234403/job/23698965959?pr=138) are expected to fail due to https://github.com/canonical/lxd/issues/13304.

tomponline (approved):
Thanks!

Can we add similar tests for the storage volume CI tests too?

roosterfish:
I have force pushed once more as the LVM thin pool test was exceeding the disk space. So the test now removes the second snapshot after every check to save disk space.

Pipeline looks clean now.

tomponline (approved):

simondeziel:
The `v1/snap1` is apparently identical to `v1/snap0` since `v1` is stopped the whole time. Should we power up `v1` in between to have different snap content to refresh? Or maybe we could grow `v1`'s rootfs in between and see if `v2` has the new size after the refresh?

@roosterfish what do you think of ^? I'd be happy to add it if that makes sense to you (I know very little about snapshot refreshes).

PR #142, opened 2024-04-16
Clone lxd-imagebuilder using git

MusicDin:
lxd-imagebuilder repository is now public so we can directly clone it instead of using deploy keys.

### DIFF
--- a/.github/actions/image-setup/action.yml
+++ b/.github/actions/image-setup/action.yml
@@ -15,13 +15,6 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Checkout LXD imagebuilder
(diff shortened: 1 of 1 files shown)


tomponline (approved):

simondeziel (approved):

simondeziel:
@MusicDin why move away from the action-checkout and manually do the clone?

MusicDin:
I find the git command a bit more straight forward, but deploy keys were easier to use with checkout action. 

Is there any benefit in using checkout action? We can switch back to it.

simondeziel:
There is no strong reason to switch back to action-checkout, except for consistency.

PR #146, opened 2024-04-16
tests/storage-volumes-vm: Add tests for custom block volume sharing

hamistao:
Adds tests for custom block volume sharing on VMs and `security.shared` usage. See canonical/lxd#13183.

### DIFF
--- a/tests/storage-volumes-vm
+++ b/tests/storage-volumes-vm
@@ -114,6 +114,20 @@ do
 	lxc storage volume attach "${poolName}" vol2 v1
 	lxc storage volume attach "${poolName}" vol3 v1
 
+	echo "==> Share a custom block volume"
+	if hasNeededAPIExtension shared_custom_block_volumes; then
+		! lxc storage volume attach "${poolName}" vol2 v2 || false
+		lxc storage volume set "${poolName}" vol2 security.shared true
+		lxc storage volume attach "${poolName}" vol2 v2
+		! lxc storage volume set "${poolName}" vol2 security.shared false || false
+		lxc storage volume detach "${poolName}" vol2 v2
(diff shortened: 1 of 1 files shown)


tests/storage-volumes-vm
```diff
@@ -99,6 +99,11 @@ do
 	lxc storage volume attach "${poolName}" vol2 v1
 	lxc storage volume attach "${poolName}" vol3 v1
 
+	echo "==> Share a custom block volume"
+	! lxc storage volume attach "${poolName}" vol2 v2 || false
+	lxc storage volume set default vol2 security.shared true
+	lxc storage volume attach "${poolName}" vol2 v2
```
roosterfish, line 105: You might also want to detach `vol2` from `v2` (and reset `security.shared`?) as part of the cleanup down below.

It doesn't really matter as both the volume and instance get removed afterwards but if you add a test case in the future it's then easier to understand what is the current state of the entities so you don't accidentally make tests with an instance or volume in a state that you don't expect it to be in.

hamistao: should I use the oportunity to detach the other volumes from `v2`?

roosterfish: You mean removing `vol5` and `vol6` which are getting attached below?
As `v2` gets removed anyway after those tests I guess its fine like this.

But attaching the same vol twice feels niche so just unsetting `security.shared` and detaching it again from v2 would feel right in my opinion. You could do this right at the end of the test you have added.
With this you make sure the test doesn't leave any traces behind.

simondeziel (commented):
Please gate this on `shared_custom_block_volumes` API extension. Otherwise, LGTM, thanks!

simondeziel (approved):

PR #147, opened 2024-04-18
Add Linux Mint

MusicDin:
Add Linux Mint daily builds and remove no longer used deploy keys.

@tomponline Secret `LXD_IMAGEBUILDER_KEY` can be removed. I guess it was not, since builds did not complain.

### DIFF
--- a/.github/workflows/image-mint.yml
+++ b/.github/workflows/image-mint.yml
@@ -1,0 +1,87 @@
+name: Build Linux Mint Images
+
+on:
+  schedule:
+    # Run at 00:00 UTC daily.
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+    inputs:
+      publish:
+        type: boolean
+        default: false
+        description: Publish built image
+
+jobs:
+  mint:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        release:
+          - una
+          - virginia
+        variant:
+          - default
+          - cloud
+        architecture:
+          - amd64
+    env:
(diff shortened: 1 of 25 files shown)


.github/workflows/image-mint.yml
```diff
@@ -0,0 +1,92 @@
...
+      fail-fast: false
+      matrix:
+        release:
+          - ulyana
+          - ulyssa
+          - uma
+          - una
+          - vanessa
+          - vera
+          - victoria
```
simondeziel, line 27: According to https://itsfoss.com/linux-mint-release-cycle/, the minor versions are not supported if a newer minor is released (20 Ulyana is not supported as folks are expected to use 20.3 Una). This feels similar to Ubuntu where 20.04 isn't supported per se, you need to use 20.04.6 (which is the latest point release).

As such, I'd keep just `una` and `virtoria`.

simondeziel (changes requested):

.github/workflows/image-mint.yml
MusicDin: I've checked on their page and it seems that all are supported at least until 2025?

https://linuxmint.com/download_all.php

MusicDin: Added only `virginia` (21.3) and `una` (20.3)

simondeziel: I guess the real test would be to compare a fully updated `20 Ulyana` and see if you get the same `/etc/os-release` data as if you used `20.3 Una`?

simondeziel (approved):

PR #149, opened 2024-04-18
github: validate yaml files

simondeziel:


simondeziel:
https://github.com/canonical/lxd-ci/actions/runs/8743923627/job/23995638438?pr=149#step:4:23 shows a bunch of warnings I didn't silence in the `.yamllint` as I think we should probably fix them at some point.

MusicDin:
Thanks 🚀 

The only notice is that `workflows/image-*.yml:3:1 [warning] truthy value should be one of [false, true] (truthy)` point on the actions's `on:` field. Can we silence those maybe?

MusicDin (approved):

tomponline (approved):

PR #153, opened 2024-04-19
images: Disable Alpine arm64 VMs

MusicDin:
Temporary disable arm64 VM builds.

Also do not build amd64 images when building arm64 images. This ensures we do not republish amd64 images when arm64 images built (especially now during testing), as it will make delta files from previous days useless.


simondeziel (approved):
I'm glad you caught the pointless `amd64` rebuild, thanks!

tomponline:
> Also do not build amd64 images when building arm64 images.

This confused me - why would we do such a thing anyway?

PR #154, opened 2024-04-19
images: Enable arm64 container builds

MusicDin:
Enable arm64 container image builds.

I've also disabled arm64 image builds for images that require cgroup v1.

### DIFF
--- a/.github/workflows/image-almalinux.yml
+++ b/.github/workflows/image-almalinux.yml
@@ -2,18 +2,22 @@ name: Build AlmaLinux Images
 
 on:
   schedule:
-    # Run at 00:00 UTC daily.
-    - cron: '0 0 * * *'
+    - cron: '0 0 * * *' # Build amd64 (Daily at 00:00 UTC).
+    - cron: '0 0 * * 1' # Build arm64 (Weekly on Monday at 00:00 UTC).
   workflow_dispatch:
     inputs:
       publish:
         type: boolean
         default: false
         description: Publish built image
+      build-arm64:
+        type: boolean
+        default: false
+        description: Build arm64 images
 
 jobs:
   almalinux:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.architecture == 'arm64' && 'Ubuntu_ARM64_4C_16G_01' || 'ubuntu-latest' }}
     strategy:
       fail-fast: false
       matrix:
@@ -25,7 +29,12 @@ jobs:
           - cloud
         architecture:
           - amd64
-          # - arm64
+          - arm64
+        build_arm64:
+          - ${{ inputs.build-arm64 == true || github.event.schedule == '0 0 * * 1' }}
+        exclude:
+          - {architecture: amd64, build_arm64: true}
+          - {architecture: arm64, build_arm64: false}
     env:
       type: "container"
       distro: "${{ github.job }}"
--- a/.github/workflows/image-centos.yml
+++ b/.github/workflows/image-centos.yml
@@ -2,18 +2,22 @@ name: Build CentOS Images
 
 on:
   schedule:
-    # Run at 00:00 UTC daily.
-    - cron: '0 0 * * *'
+    - cron: '0 0 * * *' # Build amd64 (Daily at 00:00 UTC).
+    - cron: '0 0 * * 1' # Build arm64 (Weekly on Monday at 00:00 UTC).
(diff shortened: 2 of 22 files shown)


.github/workflows/image-almalinux.yml
```diff
@@ -25,7 +29,12 @@ jobs:
           - cloud
         architecture:
           - amd64
-          # - arm64
+          - arm64
+        build_arm64:
+          - ${{ inputs.build-arm64 == true || github.event.schedule == '0 0 * * 1' }}
```
tomponline, line 34: why do we need to define the schedule twice, this looks odd.

MusicDin: The issue is that we cannot execute simple a task before the matrix is defined without creating another VM. We could potentially have only 1 schedule and run a job before the *image build* job to determine whether we should build images for arm64. But this would spawn another runner for this extra job.

So the easiest way and I found is to compare the schedule to determine whether it is arm64 or amd64 schedule.

It's not great, but seemed most efficient. I am open to suggestions if someone has a better approach?

.github/workflows/image-centos.yml
```diff
@@ -26,7 +30,13 @@ jobs:
           - cloud
         architecture:
           - amd64
-          # - arm64
+          - arm64
+        build_arm64:
+          - ${{ inputs.build-arm64 == true || github.event.schedule == '0 0 * * 1' }}
+        exclude:
+          - {architecture: amd64, build_arm64: true}
```
simondeziel, line 37: This exclude mechanism forcing a double negation is annoying but I don't see how to simplify it.

MusicDin: Yes I agree, but it is for the same reason as described in https://github.com/canonical/lxd-ci/pull/154#discussion_r1572492494.

simondeziel (approved):

.github/workflows/image-almalinux.yml
tomponline: You've lost me - but ok i trust you.

tomponline (approved):

.github/workflows/image-almalinux.yml
simondeziel: @MusicDin it might worth asking how others are tackling this issue as there are many in the same situation.

MusicDin: Let's say we have this workflow (runs everyday and builds all images):

```yaml
on:
  schedule:
    - cron: '0 0 * * *'

jobs:
  alpine:
    runs-on: ${{ matrix.architecture == 'arm64' && 'Ubuntu_ARM64_4C_16G_01' || 'ubuntu-latest' }}
    strategy:
        architecture:
          - amd64
          - arm64
```

Now if we want to build arm64 only once a week, we can add another schedule:
```yaml
on:
  schedule:
    - cron: '0 0 * * *'
    - cron: '0 0 * * 1'
```

But we need to figure out which schedule triggered the build, since we cannot provide some extra data to the event. It would be nice if pass some schedule metadata:
```yaml
on:
  schedule: # Does not work.
    - {cron: '0 0 * * *', build: 'amd64'}
    - {cron: '0 0 * * 1', build: 'arm64'}   
```

Since this does not work, we need to compare the schedules. Here is even an example in the [docs - events#schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).

---

We could have only 1 schedule (cron), but if we run any taks/command before the build job, we would spawn another runner, so this does not seem like a viable option.
```yaml
on:
  schedule:
    - cron: '0 0 * * *'
    
jobs:
  precheck: # First runner
    steps:
      - name: Is it Monday so that we build arm64?
        run: ...

  build image: # Second runner
    strategy:
      matrix:
        ...
```

Another alternative is to have multiple jobs for multiple architectures, but again, this introduces a lot of duplication.

In any case, I couldn't find a way to get rid of comparing schedules.

--

The only thing we could potentially simplify is to get rid of `exclude`s:
```yaml
jobs:
  build-image:
    strategy:
      matrix:
        ...
        architecture:
          - ${{ (inputs.build-arm64 == true || github.event.schedule == '0 0 * * 0') && 'arm64' || 'amd64' }}
```

tomponline: thanks for the explanation! appreciate it.

PR #157, opened 2024-04-23
`s/22.04/24.04`

simondeziel:
While at it, cleanup how the per-script test image is defined.


simondeziel:
@tomponline I didn't get around to figuring the DNS resolution issue with `tests/network-ovn` when using 24.04 images so I've revert to using 22.04 for that single test until I can revisit this.

I think it's still worth merging as-is.

PR #159, opened 2024-04-23
Revert "github/workflows: Disable Debian Sid"

simondeziel:
Let's try it again, assuming the `t64` transition is over with.

This reverts commit b21d51e4816a6a5d81f38a59583acf6847b18e46.

### DIFF
--- a/.github/workflows/image-debian.yml
+++ b/.github/workflows/image-debian.yml
@@ -24,7 +24,7 @@ jobs:
         release:
           - buster
           - bullseye
-          #- sid
+          - sid
           - bookworm
           - trixie
         variant:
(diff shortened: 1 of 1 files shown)


simondeziel:
@MusicDin do you remember why you had to disable `sid` back then? It's now failing due to what seems to be a `usrmerge` issue as both 22.04 and 24.04 builders fail during `debootstrap`:

```
I: Extracting zlib1g...
W: Upgrading non-merged-/usr environments post-bookworm is unsupported. Only do this for CI/QA infrastructure that will be re-bootstrapped rather than upgraded.
W: Failure trying to run: chroot "/root/build/rootfs" /bin/true
W: See /root/build/rootfs/debootstrap/debootstrap.log for details
```

```
root@t1:~/lxd-ci# ll /root/build/rootfs/bin/true
ls: cannot access '/root/build/rootfs/bin/true': No such file or directory
root@t1:~/lxd-ci# ll /root/build/rootfs/usr/bin/true
-rwxr-xr-x 1 root root 35664 Mar  9 17:42 /root/build/rootfs/usr/bin/true*
```

simondeziel:
Sid still doesn't build and now even Trixie is broken.

PR #162, opened 2024-04-25
images: Ubuntu Noble desktop

MusicDin:
Disable systemd-networkd on Ubuntu Noble desktop images which causes delays on boot.

I've left images for previous releases intact, but as @daniloegea suggested, should we disable systemd-networkd for all previous releases as well, if Ubuntu Desktop uses NetworkManager by default?

Closes https://github.com/canonical/lxd/issues/13387

### DIFF
--- a/images/ubuntu.yaml
+++ b/images/ubuntu.yaml
@@ -671,11 +671,23 @@ actions:
     #!/bin/sh
     set -eux
 
-    # Enable systemd-networkd
-    systemctl enable systemd-networkd
-
     # Disable UA attach
     systemctl mask ua-auto-attach
+
+  # Ubuntu Desktop images come with NetworkManager by default,
+  # so we should stop enabling systemd-networkd from Noble onwards.
+- trigger: post-packages
+  action: |-
+    #!/bin/sh
+    set -eux
+
+    # Enable systemd-networkd
+    systemctl enable systemd-networkd
+  releases:
+  - bionic
+  - focal
+  - jammy
+  - mantic
 
 - trigger: post-packages
   action: |-
--- a/.github/workflows/image-ubuntu.yml
+++ b/.github/workflows/image-ubuntu.yml
@@ -22,7 +22,7 @@ jobs:
           - focal
           - jammy
           - mantic
-          # - noble
+          - noble
         variant:
           - desktop
           # - default


images/ubuntu.yaml
```diff
@@ -666,16 +666,40 @@ actions:
...
-    systemctl mask ua-auto-attach
+  # Use NetworkManager from Noble release as systemd-networkd is causing
+  # delays on boot. Ubuntu Desktop images also come with NetworkManager by
+  # default, so we should start use it instead of systemd-networkd for
+  # all future releases.
+- trigger: post-packages
+  action: |-
+    #!/bin/sh
+    # Mask systemd-networkd
+    systemctl mask systemd-networkd
```
simondeziel, line 698: In line 683 we enable it so I think we should instead just not enable it to begin with?

MusicDin: Yes, but we enable it for all previous releases. I just wanted to ensure systemd-networkd is always disabled on Noble.

simondeziel:
> I've left images for previous releases intact, but as @daniloegea suggested, should we disable systemd-networkd for all previous releases as well, if Ubuntu Desktop uses NetworkManager by default?

I just did a `systemctl disable --now systemd-networkd.service` on a Jammy VM and it seems that `netplan` re-enables it at runtime on the next reboot.

If masked, it breaks the network and the Jammy desktop doesn't get any IP, despite NM also being running.

images/ubuntu.yaml
simondeziel: Ah I see. However you made the line 683 apply only for Mantic and below so at worst it's redundant/safety?

Either way, please add a link to the LP bug https://bugs.launchpad.net/subiquity/+bug/2063331 just so that we remember to remove it once fixed.

MusicDin:
Good, then we can leave older releases as is for now. I've tested only Noble and systemd-networkd remains masked and (re)boot.

simondeziel:
So on Jammy at least, it seems that NM is not strictly required but let's leave it as-is because it can be handy for folks wanting to connect to VPNs and what not. As said before, on Jammy at least, `systemd-networkd` is required so I'm fine with that workaround being for Noble+

images/ubuntu.yaml
simondeziel: I just checked and it's not enabled on Noble if we don't do it ourselves. So let's not mask it then, if that's OK for you.

```
root@nd1:~# systemctl is-enabled systemd-networkd
disabled
```

simondeziel (approved):

tomponline (approved):

PR #167, opened 2024-05-02
Fix name of `secureboot` requirement

simondeziel:
Fixes https://github.com/canonical/lxd/issues/13409

Also, fix/improve a few things along the way.

It's too bad that `bin/test-image` cannot consult the image requirements otherwise it would have been possible to look if the image being tested is not going to work with `secureboot`. For now I hardcoded the known incompatible distros but if @MusicDin you know better, please let me know!

### DIFF
--- a/bin/test-image
+++ b/bin/test-image
@@ -241,7 +241,7 @@ for url in $(lxc query "/1.0/instances" | jq -r .[] | grep "${TEST_IMAGE}"); do
 
     # DNS resolution
     DNS=0
-    for i in $(seq 3); do
+    for _ in $(seq 3); do
         if lxc exec "${name}" -- getent hosts canonical.com >/dev/null 2>&1; then
             DNS=1
             break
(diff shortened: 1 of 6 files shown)


bin/test-image
```diff
@@ -72,8 +72,14 @@ if [ "${TYPE}" = "vm" ]; then
...
         --vm \
         -c limits.cpu=4 \
-        -c limits.memory=4GiB \
-        -c security.secureboot=false
+        -c limits.memory=4GiB
+
+    # Some distros don't support secure boot.
+    case "${DISTRO}" in
+        alpine|archlinux|gentoo|nixos)
+            lxc config set "${TEST_IMAGE}" security.secureboot=true
```
tomponline, line 80: shouldnt this be false?

Whats with the `)` at the end but no starting bracket, is this some weird bash-ism?

simondeziel: Oh, indeed, `false` it should have been. Fixed, thanks.

As for the `)` missing the opening one, it is not a bashism but now that you call it out it looks weird. `man dash`:
```
The syntax of the case command is

       case word in
       [(]pattern) list ;;
       ...
       esac

The pattern can actually be one or more patterns (see Shell Patterns described later), separated by “|” characters.
The “(” character before the pattern is optional.
```
I don't remember seeing `case` with an opening `(` though. I'm find introducing it if that reads better for you.

tomponline: Ah ok thanks, from my PHP, Java, Go, C eyes thats a definite code smell, but seems like its ok in shell.

tomponline:
Thanks!

PR #171, opened 2024-05-17
Update testflinger reserve function

hamistao:
some minor changes to `bin/tf-reserve`

### DIFF
--- a/bin/tf-reserve
+++ b/bin/tf-reserve
@@ -4,6 +4,7 @@
 tf-reserve() {
     queue="${1:-anything}"
     data="${2:-jammy}"
+    lp_user="${LP_USER:-$(whoami)}"
     if echo "${data}" | grep -q ^http; then
         provision_data="url: ${data}"
     else
@@ -16,7 +17,7 @@ provision_data:
   ${provision_data}
 reserve_data:
   ssh_keys:
-    - lp:$(whoami)
+    - lp:${lp_user}
   timeout: 21600
 EOF
 }
--- a/bin/tf-reserve
+++ b/bin/tf-reserve
@@ -3,7 +3,7 @@
 # testflinger-reserve
 tf-reserve() {
     queue="${1:-anything}"
-    data="${2:-jammy}"
+    data="${2:-noble}"
     lp_user="${LP_USER:-$(whoami)}"
     if echo "${data}" | grep -q ^http; then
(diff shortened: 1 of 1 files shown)


bin/tf-reserve
```diff
@@ -3,7 +3,8 @@
 # testflinger-reserve
 tf-reserve() {
     queue="${1:-anything}"
-    data="${2:-jammy}"
+    data="${2:-noble}"
+    lp_user="${3:-$(whoami)}"
```
simondeziel, line 7: Adding `export LP_USERNAME=hamistao` to your `~/.profile` or `~/.bashrc` feels easier than always having to specify your LP username as the 3rd parameter.
```suggestion
    lp_user="${LP_USERNAME:-$(whoami)}"
```

simondeziel (approved):
See my suggestion about an env variable for the `LP_USERNAME`. Either way, that looks good to me.

Thanks for bumping to Noble.

bin/tf-reserve
hamistao: Good idea! Since some people won't be using it all the time so I think it is nice to keep the parameter, what do you think about `lp_user="${3:-${LP_USERNAME:-$(whoami)}}"`?

simondeziel: I feel that's providing too much choice and needlessly going to make it harder to ever add more positional parameters eventually.

hamistao: that's fair. In that case I think it makes sense to remove the `whoami` option because we don't have any reason to assume those would be the same values, leaving it as `${3:-LP_USERNAME}` or just `LP_USERNAME` if you think the parameter is not needed.

simondeziel: I'd prefer to leave `$3` out of this :)
For consistency, you might want to use `LP_USER` instead of the longer form.

simondeziel (approved):

simondeziel:
Now we just need to figure why all VMs test are failing (looks like GHA runner changed) but that's an issue for another PR.

simondeziel:
@tomponline please take a look and feel free to merge despite the unrelated CI failures (I'm looking into it in #173).

hamistao:
I was thinking the same, I am having similar problems with other [PR](https://github.com/canonical/lxd-ci/pull/172)

tomponline (approved):

PR #172, opened 2024-05-17
Add tests for handling disk devices with long names

hamistao:
Devices with long names can cause some problems due to socket paths or device tags that include those names being too long, resulting in errors from QEMU. See [#13320](https://github.com/canonical/lxd/pull/13320) and [#15516](https://github.com/canonical/lxd/pull/13516).

### DIFF
--- a/tests/storage-disks-vm
+++ b/tests/storage-disks-vm
@@ -109,6 +109,19 @@ waitInstanceReady v1
 lxc exec v1 -- ls -l /mnt/foo1
 lxc stop -f v1
 
+# Check adding a disk to a vm with a long name to test possible problems with long socket paths or long qemu device tags
+LONG_DEVICE_NAME="device-with-very-long-name-for-long-qemu-property-handling-test"
+
+lxc init "${IMAGE}" v2 --vm
+lxc config device add v2 "${LONG_DEVICE_NAME}" disk source="${testRoot}/allowed1" path="/mnt/bar1"
+lxc start v2
+waitInstanceReady v2
+lxc exec v2 -- mountpoint /mnt/bar
+lxc config device remove v2 "${LONG_DEVICE_NAME}"
+lxc config device add v2 "${LONG_DEVICE_NAME}" disk source="${testRoot}/allowed1" path="/mnt/bar2" # Test hotplugging as well
+lxc exec v2 -- mountpoint /mnt/bar
+lxc delete -f v2
+
 # Check adding a disk with a source path that is allowed that symlinks to another allowed source path isn't
 # allowed at start time.
 lxc config device set v1 d1 source="${testRoot}/allowed1/not-allowed2"


hamistao:
@tomponline I changed the approach of this PR from using long VM names to using long device names because that way we can check for regressions for both [#13320](https://github.com/canonical/lxd/pull/13320) and [#15516](https://github.com/canonical/lxd/pull/13516).
As a consequence, the branch name is now unrelated to the PR's changes, if this is a problem please let me know so I can close this PR and open another one with another branch name.

tests/storage-disks-vm
```diff
@@ -109,6 +109,18 @@ waitInstanceReady v1
...
+# Check adding a disk to a vm with a long name to test possible problems with long socket paths or long qemu device tags
+LONG_DEVICE_NAME="device-with-very-long-name-for-long-qemu-property-handling-tests"
+
+lxc init "${IMAGE}" v2 --vm
+lxc config device add v2 "${LONG_DEVICE_NAME}" disk source="${testRoot}/allowed1" path="/mnt"
+lxc start v2
+waitInstanceReady v2
+lxc config device remove v2 "${LONG_DEVICE_NAME}"
+lxc config device add v2 "${LONG_DEVICE_NAME}" disk source="${testRoot}/allowed1" path="/mnt" # Test hotplugging as well
+lxc exec v2 -- ls -l /mnt
```
roosterfish, line 121: I am wondering if this test is sufficient as `/mnt` is one of the directories that is always present. So even if the hotplugging fails internally the directory would still be there and the command succeeds. 

What do you think about picking something like `path=/mnt/foo` like in the test above?

roosterfish (dismissed):

tests/storage-disks-vm
```diff
@@ -109,6 +109,18 @@ waitInstanceReady v1
 lxc exec v1 -- ls -l /mnt/foo1
 lxc stop -f v1
 
+# Check adding a disk to a vm with a long name to test possible problems with long socket paths or long qemu device tags
+LONG_DEVICE_NAME="device-with-very-long-name-for-long-qemu-property-handling-tests"
+
+lxc init "${IMAGE}" v2 --vm
+lxc config device add v2 "${LONG_DEVICE_NAME}" disk source="${testRoot}/allowed1" path="/mnt/foo"
+lxc start v2
+waitInstanceReady v2
```
simondeziel, line 118: I'd check if the cold-plugged device was mounted. I'd suggest using `lxc exec ... mountpoint /mnt/foo`

tests/storage-disks-vm
```diff
@@ -109,6 +109,18 @@ waitInstanceReady v1
...
+lxc config device remove v2 "${LONG_DEVICE_NAME}"
```
simondeziel, line 119: And here make sure it's no longer mounted.

tests/storage-disks-vm
```diff
@@ -109,6 +109,18 @@ waitInstanceReady v1
...
+lxc config device add v2 "${LONG_DEVICE_NAME}" disk source="${testRoot}/allowed1" path="/mnt/foo" # Test hotplugging as well
```
simondeziel, line 120: How about using `path="/mnt/bar"`?

tests/storage-disks-vm
```diff
@@ -109,6 +109,18 @@ waitInstanceReady v1
...
+lxc exec v2 -- ls -l /mnt/foo
```
simondeziel, line 121:
```suggestion
lxc exec v2 -- mountpoint /mnt/bar
```

tests/storage-disks-vm
simondeziel: Ah sorry, I didn't review the rest of the script so if removals are tested, please ignore me. Thanks

tests/storage-disks-vm
```diff
@@ -109,6 +109,19 @@ waitInstanceReady v1
 lxc exec v1 -- ls -l /mnt/foo1
 lxc stop -f v1
 
+# Check adding a disk to a vm with a long name to test possible problems with long socket paths or long qemu device tags
+LONG_DEVICE_NAME="device-with-very-long-name-for-long-qemu-property-handling-test"
+
+lxc init "${IMAGE}" v2 --vm
+lxc config device add v2 "${LONG_DEVICE_NAME}" disk source="${testRoot}/allowed1" path="/mnt/bar"
```
simondeziel, line 116: If we take just that test case in isolation, I think it'd be better to have the cold-plugged mount path to be different than the hot-plugged one.

PR #175, opened 2024-05-23
tests/storage-vm: Check volume size rounding

MggMuggins:
Per Tom's request in https://github.com/canonical/lxd/pull/13470

EDIT: Not clear why tests are passing; they shouldn't. Will look at this more tomorrow.

### DIFF
--- a/tests/storage-vm
+++ b/tests/storage-vm
@@ -45,6 +45,12 @@ for poolDriver in $poolDriverList; do
                 [ "$(lxc storage get "${poolName}" size)" = "20GiB" ]
         fi
 
+        # Ensure non-power-of-two sizes are rounded appropriately for zfs
+        if [ "${poolDriver}" = "zfs" ]; then
+                lxc init "${IMAGE}" rounded --vm -s "${poolName}" -d root,size=13GB
+                lxc delete rounded
+        fi
+
         if [ "${poolDriver}" != "powerflex" ]; then
                 echo "==> Create VM and boot with small root (3.5GiB == 3584MiB)"
                 lxc init "${IMAGE}" v1 --vm -s "${poolName}" -d root,size=3584MiB


tomponline:
latest/edge includes your fix now, but other channels tested here do not.

Does this add weight to the idea that it is a recent kernel change not present in in GH runners (yet) that is triggering this issue?

simondeziel:
I get the same rounding behavior on:

```
root@c2d:~# lxc storage volume create default foo --type block size=24000000000
Storage volume foo created
root@c2d:~# zfs get -p volsize default/custom/default_foo
NAME                        PROPERTY  VALUE    SOURCE
default/custom/default_foo  volsize   24000004096  local
root@c2d:~# snap list lxd
Name  Version      Rev    Tracking   Publisher   Notes
lxd   git-5d43967  28646  5.21/edge  canonical✓  -
root@c2d:~# uname -a
Linux c2d 6.5.0-35-generic #35~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May  7 09:00:52 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

and:

```
sdeziel@sdeziel-lemur:~$ lxc storage volume create default foo --type block size=24000000000
Storage volume foo created
sdeziel@sdeziel-lemur:~$ zfs get -p volsize default/custom/default_foo
NAME                        PROPERTY  VALUE    SOURCE
default/custom/default_foo  volsize   24000004096  local
sdeziel@sdeziel-lemur:~$ snap list lxd
Name  Version      Rev    Tracking     Publisher   Notes
lxd   git-a2426b7  28721  latest/edge  canonical✓  -
sdeziel@sdeziel-lemur:~$ uname -a
Linux sdeziel-lemur 6.5.0-35-generic #35~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May  7 09:00:52 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
sdeziel@sdeziel-lemur:~$ modinfo zfs| grep ^version:
version:        2.2.0-0ubuntu1~23.10.2
```

tomponline:
Lxd previously rounded it to 8k

MggMuggins:
It's not the kernel...
```
wesley@jammy-azure-hwe-simple:~$ lxc init ubuntu:noble vtest --vm
Creating vtest
Error: Failed instance creation: Failed creating instance from image: Failed to run: zfs set volsize=13000007680 zfs/virtual-machines/vtest.block: exit status 255 (cannot set property for 'zfs/virtual-machines/vtest.block': 'volsize' must be a multiple of volume block size (16K))
wesley@jammy-azure-hwe-simple:~$ uname -a
Linux jammy-azure-hwe-simple 6.5.0-1021-azure #22~22.04.1-Ubuntu SMP Tue Apr 30 16:08:18 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
wesley@jammy-azure-hwe-simple:~$ sudo lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy
wesley@jammy-azure-hwe-simple:~$ snap list lxd
Name  Version        Rev    Tracking      Publisher   Notes
lxd   5.0.3-d921d2e  28373  5.0/stable/…  canonical✓  -
```
Somehow it's an inadequate test:
```
wesley@jammy-azure-hwe-simple:~$ lxc init ubuntu:noble vtest --vm
Creating vtest
Error: Failed instance creation: Failed creating instance from image: Failed to run: zfs set volsize=13000007680 zfs/virtual-machines/vtest.block: exit status 255 (cannot set property for 'zfs/virtual-machines/vtest.block': 'volsize' must be a multiple of volume block size (16K))
wesley@jammy-azure-hwe-simple:~$ lxc storage volume create zfs rt --type=block size=13GB
Storage volume rt created
wesley@jammy-azure-hwe-simple:~$ lxc storage volume show zfs custom/rt
config:
  size: 13GB
description: ""
name: rt
type: custom
used_by: []
location: none
content_type: block
project: default
created_at: 0001-01-01T00:00:00Z
```
Will look more on Tuesday.

tests/storage-vm
```diff
@@ -45,6 +45,10 @@ for poolDriver in $poolDriverList; do
                 [ "$(lxc storage get "${poolName}" size)" = "20GiB" ]
         fi
 
+        # Ensure non-power-of-two sizes are rounded appropriately (most relevant for zfs)
```
roosterfish, line 48: Can we please only execute this if the `${poolDriver}` is not `powerflex`? In PowerFlex valid volume sizes are always multiples of `8GiB`. It will fail the volume creation if other sizes like `13GB` are provided.

roosterfish: It's just that when using PowerFlex you basically cannot create a volume with sizes different than 8, 16, 24, 32, 40, .... PowerFlex's API will return an error.

So when running `lxc storage volume create "${poolName}" roundVolSize --type=block size=13GB` PowerFlex will complain.

MggMuggins: Sure... but the other drivers round their inputs silently instead of letting the backing driver error out (e.g. 16KiB ZFS block boundaries)

tomponline: The difference is that at the size of gigabytes the rounding is neglible in comparison to the allocated size, whereas with powerflex if i ask for 1gib i would actually get 8gib, which is noticeable. So we made the decision to report it to the user rather than round.

simondeziel: How much rounding should LXD do before failing? Isn't it with LVM that big'ish rounding is also needed to meet the "full PE" count?

tomponline: Certainly not Gigabytes

MggMuggins:
As much as I'd like to understand why I can only replicate the failure by creating VMs instead of just storage volumes, this is now failing as I expect it to and can be used to ensure the fix is backported.

tomponline (approved):

tomponline:
Thanks!

tests/storage-vm
```diff
@@ -45,6 +45,12 @@ for poolDriver in $poolDriverList; do
                 [ "$(lxc storage get "${poolName}" size)" = "20GiB" ]
         fi
 
+        # Ensure non-power-of-two sizes are rounded appropriately for zfs
+        if [ "${poolDriver}" = "zfs" ]; then
```
simondeziel, line 49: I would have expected:
```
        if [ "${poolDriver}" != "powerflex" ]; then
```
But maybe the rounding logic is really `zfs` specific? I thought something similar was done for all of them at the very least to be 4k-aligned (or 1MiB-aligned) and avoid the `read-write-modify` pathological case.

MggMuggins: `MinBlockBoundary` is [here](https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/utils.go#L25); the code is a bit convoluted but it should be used for dir, btrfs, zfs, and lvm. I don't see any uses in ceph or powerflex. Happy to make the change if you'd like.

simondeziel: After thinking some more, having the outer raw file/zvol/etc not be aligned to 4KiB or 1MiB shouldn't matter as the inner OS of the VM will likely have logic to do the alignment itself anyway.

Let's keep the test as is (`zfs` only) as creating a 13GB file on dir is probably too slow anyway. Thanks for digging.

PR #179, opened 2024-05-30
Add test for image conversion using lxd-migrate

MusicDin:
Adds tests for image conversion using lxd-migrate.

Tests:
- Container migration over conversion API.
- VM conversion from QCow2, and VMDK formats.
- Ensure `lxd-migrate` falls back to migration if server does not support conversion API extension (this should occur both for VMs and containers).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -86,6 +86,7 @@ jobs:
           - cluster
           - container
           - container-copy
+          - conversion
           - cpu-vm
           - devlxd-vm
           - docker
@@ -118,6 +119,8 @@ jobs:
             os: "24.04"
           # not compatible with 4.0/*
           - test: container-copy
+            track: "4.0/edge"
+          - test: conversion
             track: "4.0/edge"
           - test: cpu-vm
             track: "4.0/edge"
--- a/tests/conversion
+++ b/tests/conversion
@@ -1,0 +1,278 @@
+#!/bin/sh
+set -eu
+
+# Source helper functions.
+. ./bin/helpers
+
+install_deps jq unzip
+
+# Install LXD
+install_lxd
+
+set -x
+
+# Install go from Snap.
+snap install go --classic
+
+# Install latest lxd-migrate binary tool.
+CGO_ENABLED=0 go install github.com/canonical/lxd/lxd-migrate@latest
+
+export PATH="${HOME}/go/bin:${PATH}"
+
+# Configure LXD
+lxd init --auto --network-address "[::]" --network-port "8443"
+
+checkIPAddresses() {
+    instName=$1
+
+    address=$(lxc query "/1.0/instances/${instName}/state" | jq -r ".network.eth0.addresses | .[] | select(.scope | contains(\"global\")) | .address" 2>/dev/null || true)
+    if [ -z "${address}" ]; then
+        address=$(lxc query "/1.0/instances/${instName}/state" | jq -r ".network.enp5s0.addresses | .[] | select(.scope | contains(\"global\")) | .address" 2>/dev/null || true)
+    fi
+
+    if [ -z "${address}" ]; then
+        echo "===> FAIL: No network interface: ${instName}"
+
+        # Show the network state.
+        echo "===> DEBUG: network state: ${instName}"
+        lxc info "${instName}"
+        return 1
+    fi
+
+    fail=0
+
+    # IPv4 address
+    if echo "${address}" | grep -qF "."; then
+        echo "===> PASS: IPv4 address: ${instName}"
+    else
+        echo "===> FAIL: IPv4 address: ${instName}"
+        fail=1
+    fi
+
+    # IPv6 address
+    if echo "${address}" | grep -qF ":"; then
+        echo "===> PASS: IPv6 address: ${instName}"
+    else
+        echo "===> FAIL: IPv6 address: ${instName}"
+        fail=1
+    fi
+
+    return "${fail}"
+}
+
+checkDNS() {
+    instName=$1
+
+    # Test lxd-agent and DNS resolution.
+    if lxc exec "${instName}" -- nslookup canonical.com >/dev/null 2>&1; then
+        echo "===> PASS: DNS resolution: ${instName}"
+        return 0
+    fi
+
+    if lxc exec "${instName}" -- getent hosts canonical.com >/dev/null 2>&1; then
+        echo "===> PASS: DNS resolution: ${instName}"
+        return 0
+    fi
+
+    echo "===> FAIL: DNS resolution: ${instName}"
+    return 1
+}
+
+# conversion_vm creates a new LXD virtual machine from the image on the given path.
+# A dedicated storage pool of the given type will be created for a new virtual machine.
+#
+# Input arguments:
+# 1: Name of new virtual machine.
+# 2: Type of the storage pool.
+# 3: Path to the image.
+conversion_vm() {
+    conversion vm "$1" "$2" "$3" "$4"
+}
+
+# conversion_container creates a new LXD container from the filesystem of the existing container.
+#
+# Input arguments:
+# 1: Name of the existing container.
+conversion_container() {
+    conversion container "$1" "" "/" "no"
+}
+
+# conversion runs lxd-migrate for the given image path. For virtual machine, lxd-migrate is
+# executed on the localhost. For container, lxd-migrate will be installed and executed from
+# an existing container.
+conversion() {
+    instType=$1
+    instName=$2
+    poolType=$3
+    imgPath=$4
+    uefi_sb=$5
+
+    if [ "${instType}" = "vm" ]; then
+        echo "==> TEST: Conversion: Import virtual-machine '${instName}' on '${poolType}' storage pool"
+
+        poolName="conversion-${poolType}"
+        hostAddr="127.0.0.1"
+        instTypeCode="2" # VM in migration questioneer.
+
+        lxdMigrateCmd="lxd-migrate --conversion format"
+
+        # Create storage pool.
+        lxc storage create "${poolName}" "${poolType}"
+
+    elif [ "${instType}" = "container" ]; then
+        echo "==> TEST: Conversion: Import container '${instName}'"
+
+        poolName=""
+        hostAddr=$(lxc network get lxdbr0 ipv4.address | cut -d/ -f1)
+        instTypeCode="1" # Container.
+
+        lxdMigrateCmd="lxc exec ${instName} -- /root/go/bin/lxd-migrate"
+
+        # Install rsync and lxd-migrate.
+        lxc exec "${instName}" -- apt-get update
+        lxc exec "${instName}" -- apt-get install --no-install-recommends -y rsync file
+        lxc exec "${instName}" -- snap install go --classic
+        lxc exec "${instName}" --env CGO_ENABLED=0 -- go install github.com/canonical/lxd/lxd-migrate@latest
+
+        # Set instName to the name of the new container that will be created
+        # from the existing one.
+        instName="${instName}-migration"
+    else
+        echo "Invalid instance type '${instType}'. Valid types are 'container' and 'vm'."
+        return 1
+    fi
+
+    # Generate trust token for conversion.
+    token="$(lxc config trust add --quiet --name migrate)"
+    if [ -z "${token}" ]; then
+        echo "Failed to generate LXD trust token!"
+        return 1
+    fi
+
+    # Migration questions.
+    {
+        if [ "${instType}" = "vm" ]; then
+            echo "n"           # Do not use local unix socket.
+        fi
+
+        echo "${hostAddr}"     # Address of the target LXD server.
+        sleep 1
+        echo "y"               # Yes, this is correct fingerprint.
+        sleep 1
+        echo "1"               # Use a certificate token.
+        echo "${token}"        # Token.
+        echo "${instTypeCode}" # Instace type (1 == container, 2 == virtual-machine).
+
+        if [ "$(lxc project ls -f csv | wc -l)" -gt 1 ]; then
+            echo "default"     # Project name (required if there is more then 1 project)
+        fi
+
+        echo "${instName}"     # Instance name.
+        echo "${imgPath}"      # Local image path (or filesystem path in case of container).
+        echo "${uefi_sb}"      # Enable UEFI secure boot.
+
+        # Configure storage pool.
+        if [ "${poolName}" != "" ]; then
+            echo "4"           # Change storage pool settings.
+            echo "${poolName}" # Pool name.
+            echo "yes"         # Configure pool size?
+            echo "10GiB"       # Pool size.
+        fi
+
+        # Begin the migration with the above configuration
+        sleep 1
+        echo "1"
+    } | $lxdMigrateCmd
+
+    # Start the instance.
+    echo "Starting instance ${instName}..."
+    lxc start "${instName}"
+
+    # Wait for the instance to be ready and ensure it has global ip address.
+    echo "Waiting instance ${instName} to start..."
+    waitInstanceBooted "${instName}"
+    sleep 10
+
+    # Print the instances.
+    lxc list
+
+    echo "Check network connectivity of instance ${instName}..."
+    checkIPAddresses "${instName}"
+
+    echo "Check DNS resolution of instance ${instName}..."
+    checkDNS "${instName}"
+
+    # Cleanup.
+    lxc delete -f "${instName}"
+
+    if [ "${poolName}" != "" ]; then
+        lxc storage delete "${poolName}"
+    fi
+}
+
+# Test container migration using conversion mode. If server does not
+# support conversion API extension, lxd-migrate must fallback to
+# migration mode and successfully transfer the rootfs.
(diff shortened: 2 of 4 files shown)


MusicDin:
This test should fail until https://github.com/canonical/lxd/pull/13385 is merged.

---

@simondeziel As per our discussion, this test pulls `lxd-migrate` from go packages. We could get the `Linux` artifact that contains `lxd-migrate` and is published on each action, however, this requires a Github token. If token is not provided, the error is: `You must have the actions scope to download artifacts.`

Here is an example:
```sh
# Get URL of the latest "Linux" artifact.
artifact_url=$(curl -s "https://api.github.com/repos/canonical/lxd/actions/artifacts?name=Linux&per_page=1" \
  | jq -r --arg ARTIFACT_NAME "Linux" '.artifacts[].archive_download_url')

# Download artifact and extract pre-built lxd-migrate tool.
curl -L -H "Authorization: token ${GITHUB_TOKEN}" -o artifact.zip "${artifact_url}"

# Extract lxd-migrate binary from ZIP archive.
unzip -jo artifact.zip lxd-migrate.x86_64
mv lxd-migrate.x86_64 lxd-migrate
chmod +x lxd-migrate
```

---

Alternatively, we can build it from source.

tomponline:
> Alternatively, we can build it from source.

This sounds fine as it doesn't have any special dependencies

simondeziel:
> > Alternatively, we can build it from source.
> 
> This sounds fine as it doesn't have any special dependencies

`go install github.com/canonical/lxd/lxd-migrate@latest` feels simpler/faster to me but I have no strong opinion.

This `lxd-migrate` thing is so confusingly named I never know which tool/doc section it refers to.

@MusicDin you need to add the test here https://github.com/canonical/lxd-ci/blob/main/.github/workflows/tests.yml#L82 for it to be run on GHA.

.github/workflows/tests.yml
```diff
@@ -121,6 +122,8 @@ jobs:
             track: "4.0/edge"
           - test: cpu-vm
             track: "4.0/edge"
+          - test: conversion
+            track: "4.0/edge"
```
simondeziel, line 126: Please move that one up to keep the `test` ones alpha-sorted.

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+# Source helper functions.
+. ./bin/helpers
+
+install_deps jq unzip
+
+# Install LXD
+install_lxd
+
+# Install go from Snap.
+snap install go --channel=latest/edge --classic
```
simondeziel, line 13: Why pick the `latest/edge` channel for `go`? If the default channel works, I'd stick with it, please.

This seems to actually slow things down as it goes from 1.23 to 1.22.5:

```
go (edge) 1.23-148755a from Canonical** installed
go: downloading github.com/canonical/lxd v0.0.0-20240718114802-16be33533525
go: github.com/canonical/lxd@v0.0.0-20240718114802-16be33533525 requires go >= 1.22.4; switching to go1.22.5
go: downloading go1.22.5 (linux/amd64)
```

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+
+# Install latest lxd-migrate binary tool.
+CGO_ENABLED=0 go install github.com/canonical/lxd/lxd-migrate@latest
+
+export PATH="${HOME}/go/bin:${PATH}"
+
+# Configure LXD
+lxd init --auto
+lxc config set core.https_address :8443
```
simondeziel, line 22:
```suggestion
lxd init --auto --network-address "[::]" --network-port "8443"
```

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+        # Show the network state.
+        echo "===> DEBUG: network state: ${instName}"
+        lxc info "${instName}"
+        return 1
+    fi
+
+    fail=0
+
+    # IPv4 address
+    if echo "${address}" | grep "\." -q; then
```
simondeziel, line 44:
```suggestion
    if echo "${address}" | grep -qF "."; then
```

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+        echo "===> PASS: IPv4 address: ${instName}"
+    else
+        echo "===> FAIL: IPv4 address: ${instName}"
+        fail=1
+    fi
+
+    # IPv6 address
+    if echo "${address}" | grep ":" -q; then
```
simondeziel, line 52:
```suggestion
    if echo "${address}" | grep -qF ":"; then
```

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+        lxc storage delete "${poolName}" 2>/dev/null || true
+
+        # Create storage pool.
+        lxc storage create "${poolName}" "${poolType}"
+
+    elif [ "${instType}" = "container" ]; then
+        echo "==> TEST: Conversion: Import container '${instName}'"
+
+        poolName=""
+        hostAddr=$(lxc network show lxdbr0 | awk '/ipv4.address:/ {split($2, arr, "/"); print arr[1]}')
```
simondeziel, line 128:
```suggestion
        hostAddr="$(lxc network get lxdbr0 ipv4.address | cut -d/ -f1)"
```

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+
+        poolName="conversion-${poolType}"
+        hostAddr="127.0.0.1"
+        instTypeCode="2" # VM in migration questioneer.
+
+        lxdMigrateCmd="lxd-migrate --conversion format"
+
+        # Ensure instance and storage pool with the given name do not exist.
+        lxc delete -f "${instName}" 2>/dev/null || true
+        lxc storage delete "${poolName}" 2>/dev/null || true
```
simondeziel, line 119: Why suppress stderr and ignore errors? Is this actually needed considering the cleanup at the end of the function?

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+        hostAddr=$(lxc network show lxdbr0 | awk '/ipv4.address:/ {split($2, arr, "/"); print arr[1]}')
+        instTypeCode="1" # Container.
+
+        lxdMigrateCmd="lxc exec ${instName} -- /root/go/bin/lxd-migrate"
+
+        # Install rsync and lxd-migrate.
+        lxc exec "${instName}" -- apt-get update
+        lxc exec "${instName}" -- apt-get install rsync file
+        lxc exec "${instName}" -- snap install go --channel=latest/edge --classic
+        lxc exec "${instName}" -- go install github.com/canonical/lxd/lxd-migrate@latest
```
simondeziel, line 137: No `CGO_ENABLED=0` needed there?

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+
+        poolName=""
+        hostAddr=$(lxc network show lxdbr0 | awk '/ipv4.address:/ {split($2, arr, "/"); print arr[1]}')
+        instTypeCode="1" # Container.
+
+        lxdMigrateCmd="lxc exec ${instName} -- /root/go/bin/lxd-migrate"
+
+        # Install rsync and lxd-migrate.
+        lxc exec "${instName}" -- apt-get update
+        lxc exec "${instName}" -- apt-get install rsync file
```
simondeziel, line 135:
```suggestion
        lxc exec "${instName}" -- apt-get install --no-install-recommends -y rsync file
```

Why is `file` needed? Is this used by `lxd-migrate`?

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+        lxc exec "${instName}" -- snap install go --channel=latest/edge --classic
+        lxc exec "${instName}" -- go install github.com/canonical/lxd/lxd-migrate@latest
+        lxc exec "${instName}" -- /root/go/bin/lxd-migrate --version
```
simondeziel, line 138:
```suggestion
       ${lxdMigrateCmd}  --version
```

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+        # Set instName to the name of the new container that will be created
+        # from the existing one.
+        instName="${instName}-migration"
+    else
+        echo "Invalid instance type '${instType}'. Valid types are 'container' and 'vm'."
+        return 1
+    fi
+
+    # Generate trust token for conversion.
+    token=$(lxc config trust add --name migrate | awk 'NR==2') || exit 1
```
simondeziel, line 149:
```suggestion
    token="$(lxc config trust add --quiet --name migrate)"
```

So quoting and `--quiet` ;)

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+    if [ ! "${token}" ]; then
```
simondeziel, line 150:
```suggestion
    if [ -z "${token}" ]; then
```

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+    } | $lxdMigrateCmd
+
+    # Start the instance.
+    echo "Starting instance ${instName}..."
+    lxc start "${instName}"
+
+    # Wait for the instance to be ready and ensure it has global ip address.
+    echo "Waiting instance ${instName} to start..."
+    waitInstanceReady "${instName}"
+    sleep 15
```
simondeziel, line 197:
```suggestion
    waitInstanceBooted "${instName}"
```

Maybe?

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+    # Cleanup.
+    lxc delete -f "${instName}"
+
+    if [ "${poolName}" != "" ]; then
+        lxc storage delete "${poolName}"
+    fi
+}
+
+# Test container migration using conversion mode.
+lxc launch ubuntu-minimal:22.04 c1
```
simondeziel, line 217:
```suggestion
lxc launch ubuntu-minimal-daily:24.04 c1
```
I just checked and `ubuntu-minimal:22.04` doesn't have the `snapd` snap installed so we might as well go with the 24.04 minimal image as it's smaller than the 22.04 one.

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+set -eu
+
+# Source helper functions.
+. ./bin/helpers
+
+install_deps jq unzip
+
+# Install LXD
+install_lxd
+
```
simondeziel, line 11: `set -x` here so that we can trace the important bits. Helps a lot when reading the logs.

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+    lxc image export images:alpine/edge "${IMAGE_PATH}" --vm
+    conversion_vm vm-alpine-qcow2-zfs zfs "${IMAGE_PATH}.root"
+
+    lxc image export images:centos/9-Stream "${IMAGE_PATH}" --vm
+    conversion_vm vm-centos9-qcow2-btrfs btrfs "${IMAGE_PATH}.root"
+
+    rm "${IMAGE_PATH}" "${IMAGE_PATH}.root"
+
+    # Test VMDK image.
+    wget -O "${IMAGE_PATH}" https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-amd64.vmdk
```
simondeziel, line 240:
```suggestion
    wget --no-progress -O "${IMAGE_PATH}" https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-amd64.vmdk
```

The progress makes the log overly long.

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+        echo "${token}"        # Token.
+        echo "${instTypeCode}" # Instace type (1 == container, 2 == virtual-machine).
+
+        if [ "$(lxc project ls -f csv | wc -l)" -gt 1 ]; then
+            echo "default"     # Project name (required if there is more then 1 project)
+        fi
+
+        echo "${instName}"     # Instance name.
+        echo "${imgPath}"      # Local image path (or filesystem path in case of container).
+        echo "no"              # Enable UEFI.
```
simondeziel, line 175: It's not controlling UEFI vs CSM it seems but more UEFI Secure Boot.

Could we have that disabled only for the Alpine case (assuming that's the only one not supporting SB)?

tests/conversion
MusicDin: Nice :)

tests/conversion
tomponline: This is because of the toolchain directive. Not a fan of it personally. Because then Go downloads its own version of Go, and doesn't use the system one.

tests/conversion
MusicDin: Didn't even realize that it is missing here. However, it complains on the runner if CGO is not disabled

tests/conversion
MusicDin: Yes, `rsync` and `file` are required for lxd-migrate

tests/conversion
MusicDin: I used that for cleanup during development to ensure those resources are wiped before running the script. Have removed this no longer necessary cleanup.

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+    } | $lxdMigrateCmd
+
+    # Start the instance.
+    echo "Starting instance ${instName}..."
+    lxc start "${instName}"
+
+    # Wait for the instance to be ready and ensure it has global ip address.
+    echo "Waiting instance ${instName} to start..."
+    waitInstanceReady "${instName}"
+    sleep 10
```
simondeziel, line 194:
```suggestion
    waitInstanceBooted "${instName}"
```

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+
+    # Print the instances.
+    lxc ls
```
simondeziel, line 197:
```suggestion
    lxc list
```
Please always stick to the main action verb rather than rely on command aliases. This makes debugging/searching easier.

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+        echo "${token}"        # Token.
+        echo "${instTypeCode}" # Instace type (1 == container, 2 == virtual-machine).
+
+        if [ "$(lxc project ls -f csv | wc -l)" -gt 1 ]; then
+            echo "default"     # Project name (required if there is more then 1 project)
+        fi
+
+        echo "${instName}"     # Instance name.
+        echo "${imgPath}"      # Local image path (or filesystem path in case of container).
+        echo "${uefi}"         # Enable UEFI.
```
simondeziel, line 172:
```suggestion
        echo "${uefi_sb}"      # Enable UEFI Secure Boot.
```

tests/conversion
MusicDin: Even when waitInstanceBooted is used, I still have to wait for a few seconds, otherwise the IP address may not be reported by the instance yet.

simondeziel: Ack, but still, since you need network connectivity, `waitInstanceBooted` is better, even if you add a `sleep` after it.
`waitInstanceReady` is more for `lxc exec` readyness than anything.

MusicDin:
@tomponline Ready for review/merge. Have rebased and addressed all the issues. All tests are also nicely passing.

tests/conversion
tomponline: @MusicDin is this addressed?

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+
+    # Cleanup.
+    lxc delete -f "${instName}"
+
+    if [ "${poolName}" != "" ]; then
+        lxc storage delete "${poolName}"
+    fi
+}
+
+# Test container migration using conversion mode.
```
tomponline, line 216: Why is this check not wrapped in the hasNeededAPIExtension check below, yet it mentions "conversion mode"?

MusicDin: Because `lxd-migrate` has to fallback to migration if conversion is not available. For virtual machines it will fallback as well, but the error for "unsupported" format will be thrown, therefore, I've limited it to run VM tests only when conversion is enabled, as we are not testing raw images

tomponline: Could you add a comment explaining that, because "conversion mode" is unclear here.

tests/conversion
```diff
@@ -0,0 +1,264 @@
...
+    lxc list
+
+    echo "Check network connectivity of instance ${instName}..."
+    checkIPAddresses "${instName}"
+
+    echo "Check DNS resolution of instance ${instName}..."
+    checkDNS "${instName}"
+
+    echo "Stopping instance ${instName}..."
+    lxc stop "${instName}" --timeout=30
```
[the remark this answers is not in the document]
MusicDin: Just wanted to get a clean stop, but can use `delete -f` as well.

tests/conversion
tomponline: Should this be?

```
lxc exec "${instName}" --env=CGO_ENABLED=0 -- go install github.com/canonical/lxd/lxd-migrate@latest
```

tests/conversion
```diff
@@ -0,0 +1,263 @@
...
+
+    # Cleanup.
+    lxc delete -f "${instName}"
+
+    if [ "${poolName}" != "" ]; then
+        lxc storage delete "${poolName}"
+    fi
+}
+
+# Test container migration using conversion mode. If server does not
```
tomponline, line 213: Can we also test VM raw import here too?

MusicDin:
This should fail on `5.21` and `5.0` until https://github.com/canonical/lxd/pull/13816 is merged.

tests/conversion
MusicDin: Fixed.

tomponline (approved):
Thanks!

PR #183, opened 2024-06-04
tests/storage-vm: add test for snap zfs.external

simondeziel:
Test for https://github.com/canonical/lxd-pkg-snap/pull/413

### DIFF
--- a/tests/storage-buckets
+++ b/tests/storage-buckets
@@ -22,7 +22,7 @@ elif [ "${arch}" = "ppc64el" ]; then
 elif [ "${arch}" = "armhf" ]; then
   arch="arm"
 elif [ "${arch}" = "riscv64" ] ; then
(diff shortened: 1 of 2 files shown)


simondeziel:
@tomponline this was tested locally. The current CI failures are unrelated as the test bits are inactive when `snap info lxd` doesn't contain `zfs.external`.

Now that I talk of `snap info`, I'm reminded that this might not be looking at the current data from the snap but might be checking the store's copy instead. Should I use a different gating mechanism?

simondeziel:
> Now that I talk of `snap info`, I'm reminded that this might not be looking at the current data from the snap but might be checking the store's copy instead. Should I use a different gating mechanism?

I changed it to instead look at the `/snap/lxd/current/commands/daemon.start` script directly. Not super clean but should be reliable.

PR #186, opened 2024-06-10
tests: Add basic IP allocation tests

markylaing:
I've added some basic tests for automatic allocation of IP addresses for OVN network forwards and load balancers.

I've found it difficult to make these tests more extensive because the allocated IPs are random but the tests rely on hardcoded IPs. To avoid collisions we need to keep track of whether or not forwards/load balancers have been created with allocated IPs (by the API extension) and delete them before creating new forwards/load balancers with a fixed IP on any network using the same uplink.

I think it's fine for now to just ensure they can be created/deleted and that new ones can't be created with the same IP. There is no difference in code path/functionality between allocated/fixed IP after initial creation.

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -629,10 +629,23 @@ ovn_forward_tests() {
     lxc network forward create ovn-virtual-network 2001:db8:1:2::1
     ovn-nbctl list load_balancer | grep -cF name | grep -xF 0
 
+    allocated_ipv4_address=''
+    allocated_ipv6_address=''
+    if hasNeededAPIExtension network_allocate_external_ips; then
+      echo "==> Check forwards can be created with an allocated IP address"
+      allocated_ipv4_address="$(lxc network forward create ovn-virtual-network --allocate=ipv4 | cut -d' ' -f3)"
+      allocated_ipv6_address="$(lxc network forward create ovn-virtual-network --allocate=ipv6 | cut -d' ' -f3)"
+    fi
+
     if hasNeededAPIExtension network_load_balancer; then
       echo "==> Check load balancer cannot be created on same listen address as forward"
       ! lxc network load-balancer create ovn-virtual-network 192.0.2.1 || false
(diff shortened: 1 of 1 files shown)


markylaing:
Failing here: https://github.com/canonical/lxd-ci/actions/runs/9447367697/job/26019039138?pr=186#step:7:1615

This should pass when edge contains https://github.com/canonical/lxd/pull/13575

tests/network-ovn
```diff
@@ -643,9 +656,18 @@ ovn_forward_tests() {
     echo "==> Check basic forward default target address can be set (to non-existent target address)"
     lxc network forward set ovn-virtual-network 192.0.2.1 target_address="${ip4AddressPrefix}.2"
     lxc network forward set ovn-virtual-network 2001:db8:1:2::1  target_address="${ip6AddressPrefix}::2"
-    ovn-nbctl list load_balancer | grep -cF name | grep -xF 2
+    if hasNeededAPIExtension network_allocate_external_ips; then
+      echo "==> Check basic forward default target address can be set on allocated forward (to non-existent target address)"
+      lxc network forward set ovn-virtual-network "${allocated_ipv4_address}" target_address="${ip4AddressPrefix}.3"
+      lxc network forward set ovn-virtual-network "${allocated_ipv6_address}"  target_address="${ip6AddressPrefix}::3"
+    fi
+    ovn-nbctl list load_balancer | grep -cF name | grep -xF 4
```
simondeziel, line 664: Should this `ovn-nbctl list load_balancer` be in the `if`? It seems to explain why `5.0/edge` failed: https://github.com/canonical/lxd-ci/actions/runs/9447367697/job/26019032733?pr=186#step:7:1658

markylaing: Ah yes there should be 4 in the if block and 2 otherwise. Will change.

markylaing: Fixed.

tomponline (approved):

PR #187, opened 2024-06-10
bin: Allow sideloading of lxd binaries

markylaing:
Adds a new environment variable `ENABLE_SIDELOADING` which stops the LXD daemon, then starts it again when the user presses enter. This allows time for loading a custom build.

I found this useful in my local testing.

### DIFF
--- a/bin/helpers
+++ b/bin/helpers
@@ -128,6 +128,12 @@ install_lxd() (
     # on first invocation
     mkdir -p ~/snap/lxd/common/config/
     touch ~/snap/lxd/common/config/config.yml
+
+    if [ -n "${LXD_SIDELOAD_PATH:-}" ]; then
+        systemctl stop snap.lxd.daemon
+        cp "${LXD_SIDELOAD_PATH}" /var/snap/lxd/common/lxd.debug
+        systemctl start snap.lxd.daemon
+    fi
 )
 
 # hasNeededAPIExtension: check if LXD supports the needed extension.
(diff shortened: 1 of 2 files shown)


tomponline:
> Adds a new environment variable `ENABLE_SIDELOADING` which stops the LXD daemon, then starts it again when the user presses enter. This allows time for loading a custom build.
> 
> I found this useful in my local testing.

Good idea. But how about use LXD_SIDELOAD_PATH and update the scripts to copy that path binary into the correct location and reload LXD, avoiding the need for manual actions?

markylaing:
> > Adds a new environment variable `ENABLE_SIDELOADING` which stops the LXD daemon, then starts it again when the user presses enter. This allows time for loading a custom build.
> > I found this useful in my local testing.
> 
> Good idea. But how about use LXD_SIDELOAD_PATH and update the scripts to copy that path binary into the correct location and reload LXD, avoiding the need for manual actions?

Yep sounds good. Will update.

markylaing:
Have updated to use `LXD_SIDELOAD_PATH`.

bin/helpers
```diff
@@ -128,6 +128,12 @@ install_lxd() (
     # on first invocation
     mkdir -p ~/snap/lxd/common/config/
     touch ~/snap/lxd/common/config/config.yml
+
+    if [ -n "${LXD_SIDELOAD_PATH:-}" ]; then
+        systemctl stop snap.lxd.daemon
+        cp "${LXD_SIDELOAD_PATH}" /var/snap/lxd/common/lxd.debug
+        systemctl start snap.lxd.daemon
+    fi
```
simondeziel, line 136: I'd propose doing this but before the `waitready` earlier in that function. In the microcloud tests, we do it slightly differently so maybe you'd like to reuse it, https://github.com/canonical/microcloud/blob/main/test/includes/microcloud.sh#L136-L142

tomponline: you can also just do cp, followed by reload rather than stop and start

PR #188, opened 2024-06-11
CI tweaks

simondeziel:
@tomponline I tried adding a `sleep 3` prior to checking `taskset` output but that didn't fair any better. @mihalicyn ideally the `lxc config set <instance> limits.cpu=X` should be waiting for the CPU affinity changes to be done before returning.

@tomponline this should improve the scheduling of the tests and avoid scheduled runs from cancelling manually triggered ones.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,7 +28,8 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


simondeziel:
@tomponline you are probably aware of it but those tests showed that `5.21/edge` is missing the fix for:

```
 ==> Check forwards can be created with no target address
+ lxc network forward create ovn-virtual-network 192.0.2.1
Error: Received invalid IP "192.0.2.1?project=testovn"
```

among other things.

tomponline (approved):

PR #189, opened 2024-06-11
cpu-vm: relax cpu pinning check a bit

mihalicyn:
It was discovered, that current check, while being correct in theory works really bad in practice. Because QEMU sometimes spawns additional temporary threads from an actual vCPU threads which leads to the situation when we have more threads with pinning applied than vCPU number. Let's relax our check a bit and check for "greater than" and not for the equality.

https://warthogs.atlassian.net/browse/LXD-1200


simondeziel (approved):
LGTM and makes some sense. It could do with a small comment in the actual test script. the one you have in the PR description is good IMHO.

tomponline (approved):

tomponline (approved):
Thanks!

PR #191, opened 2024-06-12
tests/cluster: switch to -minimal image

simondeziel:
This replaces `ping` that's not on -minimal images by `bash` TCP probing of SSH.

This was tested on a 24.04 host with `6.8.0-38-generic` kernel (the one re-introducing FAN support on Noble).


tomponline:
Does this have a timeout though?

tomponline (approved):

PR #192, opened 2024-06-13
test/tpm-vm: Fix TPM test so that it doesn't always fail

tomponline:
The test was missing FAIL=0 so could have never possibly been run successfully.


hamistao (approved):
LGTM

PR #202, opened 2024-06-19
Remove unecessary image source flag when building Debian images

MusicDin:


tomponline (commented):
Why is it unnecessary now?

MusicDin:
It was previously overwritten with `ftp.us.debian.org`, but now we are forcing the one that is used by default. This is just a cosmetic change to keep the invocation as minimal as possible. Can also be closed as no real change will be made.

This was changed in https://github.com/canonical/lxd-ci/commit/53dbfc01329f38fd275ab3ab5bf40eef45d86cfa

And this is the default url:

https://github.com/canonical/lxd-ci/blob/main/images/debian.yaml#L14

tomponline (approved):

PR #206, opened 2024-06-21
github: extent test matrix to 4.0/edge and 20.04

simondeziel:
This adds 15 tests to the existing test matrix. This put the current total at 105 jobs taking ~15 hours of CI runtime.

### DIFF
--- a/tests/vm-nesting
+++ b/tests/vm-nesting
@@ -94,7 +94,8 @@ start 5
 wait 5
 cmd 5 "snap wait system seed.loaded && snap install lxd --channel ${LXD_SNAP_CHANNEL}"
 cmd 5 "lxd init --auto"
-cmd 5 "lxc launch ${IMAGE} nested --vm -c limits.memory=512MiB -d root,size=3584MiB"
+cmd 5 "lxc profile device set default root size=3584MiB"
+cmd 5 "lxc launch ${IMAGE} nested --vm -c limits.memory=512MiB"
 delete 5
 
 echo "==> Test 5 containers each with one nested VM"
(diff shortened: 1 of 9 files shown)


tests/vm-nesting
```diff
@@ -94,7 +94,8 @@ start 5
 wait 5
 cmd 5 "snap wait system seed.loaded && snap install lxd --channel ${LXD_SNAP_CHANNEL}"
 cmd 5 "lxd init --auto"
-cmd 5 "lxc launch ${IMAGE} nested --vm -c limits.memory=512MiB -d root,size=3584MiB"
+cmd 5 "lxc profile device set default root size=3584MiB"
```
tomponline: could use `lxc config device override` here to mirror what `-d` was doing and avoid modifying the profile.

tomponline (approved):
Thanks!

tests/vm-nesting
simondeziel: I think I tried that at first but it refused to shrink the rootfs, probably as a protection for FS not supporting shrinking, alto that was with the regular `ext4`.

tomponline: yeah so you can do `lxc init`, `lxc config device override`, `lxc start`.
The same limitations will apply to `lxc profile device set`, it will just skip over ones that cant be skipped.

simondeziel: I presume shrinking is accepted during creation due to not putting any real data at risk. The `init + override` case could probably be adapted as the instance was never started to begin with.

simondeziel: No problem, thanks for bringing that up. Considering that nowadays, it's easy to use `-d root,size=`, I don't think it's worth trying to handle the block shrinking case post `init` but pre-`start`.

PR #208, opened 2024-06-25
Use shorter dev names on older LXD releases

simondeziel:


tomponline (dismissed):

simondeziel:
@tomponline the detection works but the `storage-disks-vm` test fails just a [bit later](https://github.com/canonical/lxd-ci/actions/runs/9667916011/job/26670827580?pr=208#step:7:377):

```
+ echo '::warning::5.21/edge detected, using a shorter name'
Warning: 5.21/edge detected, using a shorter name
+ LONG_DEVICE_NAME=notSoLongName
+ lxc init ubuntu-minimal-daily:24.04 v2 --vm
Creating v2
+ lxc config device add v2 notSoLongName disk source=/opt/lxd-test-restricted/allowed1 path=/mnt/bar1
Device notSoLongName added to v2
+ lxc start v2
+ waitInstanceReady v2
+ set +x
+ lxc exec v2 -- mountpoint /mnt/bar1
/mnt/bar1 is a mountpoint
+ lxc config device remove v2 notSoLongName
Error: Failed to stop device "notSoLongName": Failed to detach path device after 10s: Chardev 'lxd_notSoLongName' is busy
+ cleanup
Test failed
```

simondeziel:
Should I convert the hot-unplug by a cold unplug when working around?

simondeziel:
@tomponline I ended up disabling all the disk hot plug/unplug part of the storage test which I think is OK for a temporary workaround.

It's currently as green as it gets considering the ZFS vol activation regression that should be addressed by https://github.com/canonical/lxd/pull/13662. There's also the ZFS rounding issue that's still not completely backported all the way down to `5.0/edge`.

tomponline:
> I ended up disabling all the disk hot plug/unplug part of the storage test which I think is OK for a temporary workaround.

Can we keep hotplug tests? I don't think we will ever back port the fix to 5.0

tomponline:
> There's also the ZFS rounding issue that's still not completely backported all the way down to `5.0/edge`.

It should be now since https://github.com/canonical/lxd/pull/13659

simondeziel:
> > I ended up disabling all the disk hot plug/unplug part of the storage test which I think is OK for a temporary workaround.
> 
> Can we keep hotplug part of tests? Just not hot unplug, i don't think we will ever back port the fix to 5.0

Sure, done.

tomponline (approved):

PR #217, opened 2024-06-28
bin/test-image: Disable secureboot for specific images

MusicDin:
Revert https://github.com/canonical/lxd-ci/pull/205

We cannot extract `requirements` from manually built images, as LXD retrieves them from the image server - they are not part of the image metadata.

This should fix issues with VM tests on Alpine, Arch, ...

Example passing build for Alpine: https://github.com/MusicDin/lxd-ci/actions/runs/9712230286

### DIFF
--- a/bin/test-image
+++ b/bin/test-image
@@ -75,9 +75,11 @@ if [ "${TYPE}" = "vm" ]; then
         -c limits.memory=4GiB
 
     # Some distros don't support secure boot.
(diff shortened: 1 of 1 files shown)


tomponline (approved):

simondeziel:
> We cannot extract `requirements` from manually built images, as LXD retrieves them from the image server - they are not part of the image metadata.

Oh that's unfortunate, sorry for introducing breakage and thanks for catching it.

PR #218, opened 2024-06-28
Fix yaml linting issues

MusicDin:
Fixes linter issues in `.github/workflows` (inline comment spacing and identation)


tomponline (approved):
Thanks!

PR #219, opened 2024-06-28
Add `/dev/disk/by-id` link tests

hamistao:
This adds tests for checking if `/dev/disk/by-id` links are being properly set up for different kinds of block devices. Also moves non pool specific tests away from `storage-vm`.
This is needed for [#13672](https://github.com/canonical/lxd/pull/13672) and [#13671](https://github.com/canonical/lxd/pull/13671).

### DIFF
--- a/tests/storage-disks-vm
+++ b/tests/storage-disks-vm
@@ -110,8 +110,26 @@ lxc exec v1 -- ls -l /mnt/foo1
 lxc stop -f v1
 
 # Check adding a disk to a vm with a long name (max 63) and slash to test possible problems with long socket paths or long qemu device tags
-LONG_DEVICE_NAME="device-with-very-long-name-and-/-4-qemu-property-handling-test_"
+LONG_DEVICE_NAME="device-/-with-very-long-name-and-4-qemu-property-handling-test_"
 DEVICE_HOTPLUG="1"
+
+lxc launch "${IMAGE}" v2 --vm
+
+echo "==> Check /dev/disk/by-id with escaped long device name"
+# Create block device source inside allowed directory
+touch "${testRoot}/allowed1/lxd-block-test"
+truncate -s 5m "${testRoot}/allowed1/lxd-block-test"
+
+waitInstanceReady v2
+lxc config device add v2 ${LONG_DEVICE_NAME} disk source="${testRoot}/allowed1/lxd-block-test"
+sleep 3
+BLOCK_SPECIAL_FILE="$(lxc exec v2 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_device-----with-)"
+lxc exec v2 -- test -b "$BLOCK_SPECIAL_FILE"
+lxc config device remove v2 ${LONG_DEVICE_NAME}
+sleep 3
+! lxc exec v2 -- stat "$BLOCK_SPECIAL_FILE" || false
+! lxc exec v2 -- stat /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_device-----with- || false
+lxc stop -f v2
 
 # XXX: LXD releases 5.21 and earlier don't support long names (yet)
 if echo "${LXD_SNAP_CHANNEL}" | grep -E '^([45]\.0|5\.21)/'; then
@@ -124,7 +142,6 @@ if echo "${LXD_SNAP_CHANNEL}" | grep -E '^[45]\.0/'; then
   DEVICE_HOTPLUG="0"
 fi
 
-lxc init "${IMAGE}" v2 --vm
 lxc config device add v2 "${LONG_DEVICE_NAME}" disk source="${testRoot}/allowed1" path="/mnt/bar1"
 lxc start v2
 waitInstanceReady v2
@@ -230,6 +247,7 @@ lxc network delete lxdbr0
 
 rm "${testRoot}/allowed1/not-allowed1"
 rm "${testRoot}/allowed1/not-allowed2"
+rm "${testRoot}/allowed1/lxd-block-test"
 rmdir "${testRoot}/allowed1/foo1"
 rmdir "${testRoot}/allowed1/foo2"
 rm "${testRoot}/allowed1/foolink"
--- a/tests/storage-vm
+++ b/tests/storage-vm
@@ -63,10 +63,10 @@ for poolDriver in $poolDriverList; do
         lxc info v1
 
         echo "==> Check /dev/disk/by-id"
-        lxc exec v1 -- test -e /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root
-        lxc exec v1 -- test -e /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root-part1
+        lxc exec v1 -- test -b /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root
+        lxc exec v1 -- test -b /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root-part1
         if lxc exec v1 -- mount | grep -qwF /boot/efi; then
-            lxc exec v1 -- test -e /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root-part15
+                lxc exec v1 -- test -b /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root-part15
         fi
 
         echo "==> Check config drive is readonly"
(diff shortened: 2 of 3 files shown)


tests/storage-vm
```diff
@@ -301,6 +301,16 @@ for poolDriver in $poolDriverList; do
         lxc start v1
         waitInstanceReady v1
 
+        echo "==> Checking /dev/disk/by-id representations are correct"
+        lxc exec v1 -- stat -c "%F" /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro | grep -xF "symbolic link"
+	lxc exec v1 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro | grep -xF "/dev/sdb"
```
simondeziel, line 306: IIRC, we want to use `/dev/disk/by-id` to avoid depending on automatically allocated `/dev/sdX` style as the later are not guarantied to be stable.

tests/storage-vm
```diff
@@ -301,6 +301,16 @@ for poolDriver in $poolDriverList; do
         lxc start v1
         waitInstanceReady v1
 
+        echo "==> Checking /dev/disk/by-id representations are correct"
+        lxc exec v1 -- stat -c "%F" /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro | grep -xF "symbolic link"
```
simondeziel, line 305:
```suggestion
        [ "$(lxc exec v1 -- stat -c "%F" /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro)" = "symbolic link" ]
```

This is faster and easier to read IMHO in logs.

tests/storage-vm
hamistao: I figured guaranteeing the link in properly set up took care of the problem, would it be better to just substitute every `/dev/sd*` reference?

tests/storage-volumes-vm
```diff
@@ -118,22 +129,22 @@ do
 	waitInstanceReady v2
 
 	# shellcheck disable=2016
-	lxc exec v1 -- /bin/sh -c 'mount /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block-dev1 /mnt && [ $(cat /mnt/bar) = foo ] && umount /mnt'
+	lxc exec v1 -- /bin/sh -c 'mount /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vol2 /mnt && [ $(cat /mnt/bar) = foo ] && umount /mnt'
```
tomponline, line 132: These lines have changed multiple times in the same PR - please dont do this.

Each commit in this PR should be the final version of that logical change.

Additionally these line changes do not relate to the commit message:

>  tests/storage-volumes-vm: Check links to /dev/s*

tomponline (changes requested):
There seems to be some commits that are iterating on changes of the same lines - please dont do this. As discussed before each line change should be the final version of a logical change - not an iteration.

Secondly, we agreed on Friday that we would move the "Checking additional disk device support" tests from storage-vm to storage-disks-vm so as to keep the external disks together in one file and so as to avoid repeating them for each pool driver unnecessarily, however I do not see this change.

tomponline:
@hamistao additionally please can you also add some tests to the storage-disks-vm file that checks for devices passed with long & escaped names that are longer than what udev uses, so we can check the udev /dev/disk/by-id path is unchanged. Ta

hamistao:
@tomponline I am planning to include the discussed moving of the tests on a later PR along with other similar changes to make this simpler.
Also, using long names don't serve a lot of purpose here because the name is trimmed to about 16 characters when added to the `/dev/disk/by-id` path, if you think it still makes sense to use them I can make the changes. As for escaped names, I believe I am using one.

tests/storage-volumes-vm
hamistao: Sorry, this one was an accident during the last rebase, fixed

hamistao:
@tomponline The new long name tests were added, feel free to take a look at this when you can. I believe this is failing the tests due to unrelated problems with `latest/edge`.

tomponline:
Thanks, unfortunately as latest/edge is now broken ill hold off on this until @mihalicyn and I can unblock it.

tomponline:
> The new long name tests were added

You mentioned some devices didn't work with long names - please can you elaborate on this.

hamistao:
@tomponline It seemed like using long names could break the devices but in fact, the paths on `/dev/` and `/dev/disk/by-id` behave differently. For example, on `ubunt-minimal` VMs adding two devices with names that share a 16 character prefix, the `/dev/disk/by-id` path of the first would be overwritten by the second, leaving the first without a link on `/dev/disk/by-id`.
For the purpose of these tests, thay are not affected by this behavior.

hamistao:
> Thanks, unfortunately as latest/edge is now broken ill hold off on this until @mihalicyn and I can unblock it.
No problem!

tests/storage-disks-vm
```diff
@@ -211,14 +154,14 @@ if lxc exec v1 -- mount | grep -qwF /boot/efi; then
...
-  lxc stop -f v1
-  lxc config device remove v1 d1
-  lxc project unset restricted restricted.devices.disk
-  lxc project unset restricted restricted.devices.disk.paths
-  lxc project unset restricted restricted
+# Remove disk device restrictions
+lxc stop -f v1
+lxc config device remove v1 d1
+lxc project unset restricted restricted.devices.disk
+lxc project unset restricted restricted.devices.disk.paths
```
tomponline, line 161: Why do we need to do this (your commit message didnt explain the why, only the what)?

hamistao:
@tomponline I will push new changes removing the tests that use `readlink` to check links to `/dev/sd*`, as was discussed earlier with @simondeziel

tests/storage-disks-vm
tomponline: Why did the previous form of the tests not require this stuff to be turned off?

tomponline:
Good news, the latest/edge is unblocked and the tests are green.

Just need to understand about the restrictions being relaxed and then we can merge.

tests/storage-disks-vm
hamistao: this became a requirement after adding new long name tests in the previous commit (this time with a block device that uses `loopdev`). Please feel free to suggest alternative ways to perform this tests if you would like.

hamistao: yes, I just figured it made sense to keep them together as they are unrelated to the restriction tests

tests/storage-disks-vm
```diff
@@ -221,8 +175,67 @@ else
...
+# Check adding a disk to a vm with a long name (max 63) and slash to test possible problems with long socket paths or long qemu device tags
+LONG_DEVICE_NAME="device-/-with-very-long-name-and-4-qemu-property-handling-test_"
+DEVICE_HOTPLUG="1"
+
+lxc launch "${IMAGE}" v2 --vm
+
+echo "==> Check /dev/disk/by-id with escaped long device name"
+waitInstanceReady v2
+lxc config device add v2 ${LONG_DEVICE_NAME} disk source="${loopdev}"
+[ "$(lxc exec v2 -- stat -c "%F" /dev/sdb)" = "block special file" ]
```
simondeziel, line 188: This is still hardcoding `/dev/sdX` which is what we should avoid. I'd re-order the test to have a `readlink` call to know which `/dev/sdX` it points to and then use that variable later on.

simondeziel (dismissed):
Please remove remaining traces of `/dev/sdX` everywhere.

tests/storage-disks-vm
hamistao: I was just making this exact change :). Will be pushing it shortly, thanks!

hamistao: I was just making this exact change :). Will be pushing it shortly, thanks!

tests/storage-vm
```diff
@@ -343,13 +343,13 @@ for poolDriver in $poolDriverList; do
         lxc exec v1 -- rm /srv/rw/lxd-test-rw
         lxc exec v1 -- rm /srv/rw/lxd-test
 
-        # Check block disks are available.
-        lxc exec v1 -- stat -c "%F" /dev/sdb | grep -xF "block special file"
-        lxc exec v1 -- stat -c "%F" /dev/sdc | grep -xF "block special file"
+        # Check /dev/disk/by-id/ links
+        lxc exec v1 -- test -b "$(lxc exec v1 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro)"
```
simondeziel, line 347: I don't think we need to do the symlink deference manually, do we?

```suggestion
        lxc exec v1 -- test -b /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro
```

hamistao: fixed, thanks for the insight!

tests/storage-vm
```diff
@@ -378,10 +378,12 @@ for poolDriver in $poolDriverList; do
         fi
         lxc storage volume attach "${poolName}" vol1 v1
         sleep 3
-        lxc exec v1 -- stat -c "%F" /dev/sdb | grep -xF "block special file"
+        BLOCK_SPECIAL_FILE="$(lxc exec v1 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vol1)"
+	lxc exec v1 -- test -b "$BLOCK_SPECIAL_FILE"
```
tomponline, line 382: Messy spacing

tests/storage-disks-vm
```diff
@@ -234,8 +175,69 @@ else
   echo 'Skipping NVME test due to missing extension: "disk_io_bus"'
 fi
 
+
+# Check adding a disk to a vm with a long name (max 63) and slash to test possible problems with long socket paths or long qemu device tags
```
tomponline, line 179: I would prefer to keep these tests in a place where the project restrictions were still enabled so we are testing that this works with project restrictions (as we are now).

tests/storage-disks-vm
```diff
@@ -110,9 +110,23 @@ lxc exec v1 -- ls -l /mnt/foo1
 lxc stop -f v1
 
 # Check adding a disk to a vm with a long name (max 63) and slash to test possible problems with long socket paths or long qemu device tags
-LONG_DEVICE_NAME="device-with-very-long-name-and-/-4-qemu-property-handling-test_"
+LONG_DEVICE_NAME="device-/-with-very-long-name-and-4-qemu-property-handling-test_"
```
tomponline: What was the reason for the / change here?

hamistao: The reason was to make `/` show on the `/dev/disk/by-id` path so we can check for escaping

tomponline (approved):
Thanks!

PR #230, opened 2024-07-04
github: prevent accidental usage of `images:` in CI test jobs

simondeziel:


.github/workflows/tests.yml
```diff
@@ -214,6 +214,11 @@ jobs:
       - name: ${{ matrix.test }} (${{ matrix.track }})
         run: |
           set -eux
+
+          # XXX: prevent accidental usage of `images:` in CI test jobs.
```
tomponline, line 218: @MusicDin are you planning to use images remote for your virtv2v tests?

simondeziel: If/when it's something we need, I'd prefer we remove the `/etc/hosts` alias on a case by case basis, so in the specific test needing the DNS name to work.

MusicDin: May I ask for the reason why `images:` remote should not be used? Just curios

> are you planning to use images remote for your virtv2v tests?

I'm copying images here:

https://github.com/canonical/lxd-ci/pull/179/files#diff-fa5a6ef4846304f7bfee1bb83a62c735a35ee164d2d89da0fbde17e69e6b9183R215-R219

MusicDin: Ahh, I see in the commit description that only official images should be used for tests

simondeziel: For this new `conversion` test, please add a `sudo sed -i '/images.lxd.canonical.com/d' /etc/hosts` prior to making use of the `images:` remote.

MusicDin (approved):
Thanks!

tomponline (approved):

PR #231, opened 2024-07-04
Storage: Move misplaced tests

hamistao:
Moves tests away from `storage-vm`. Volume related tests were moved to `storage-volumes-vm` and pool agnostic tests to `storage-disks-vm`.
Also switches `grep -x` to `=` on `storage-vm` and removes some leftover references to `sdb.`


simondeziel:
Something's wrong with this PR as it wants to include way more commits than it should.

hamistao:
@simondeziel It was a mistake on my end, fixed.

tests/storage-vm
```diff
@@ -457,7 +322,7 @@ for poolDriver in $poolDriverList; do
 
                 echo "==> Checking VM config disk filesystem is not XFS"
                 serverPID="$(lxc query /1.0 | jq .environment.server_pid)"
-                nsenter -m -t "${serverPID}" findmnt --noheadings --output=FSTYPE --mountpoint /var/snap/lxd/common/lxd/devices/v1/config.mount | grep -xF ext4
+                [ "$(nsenter -m -t "${serverPID}" findmnt --noheadings --output=FSTYPE --mountpoint /var/snap/lxd/common/lxd/devices/v1/config.mount)" = ext4 ]
```
simondeziel, line 325: As a matter of style, please quote the right hand side of the comparison as well. Please also do so for all new occurrences, thanks!

tests/storage-volumes-vm
```diff
@@ -69,6 +69,18 @@ do
...
+	lxc start v1
+	waitInstanceReady v1
+	lxc storage volume attach "${poolName}" vol1 v1
+	sleep 3
+	BLOCK_SPECIAL_FILE="$(lxc exec v1 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vol1)"
+	lxc exec v1 -- test -b "$BLOCK_SPECIAL_FILE"
+	lxc storage volume detach "${poolName}" vol1 v1
+	sleep 3
+	! lxc exec v1 -- stat "$BLOCK_SPECIAL_FILE" || false
+	! lxc exec v1 -- stat /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_vol1 || false
```
simondeziel, line 82: I'd prefer `s/stat/test -e/` as it more directly convey the intention.

tests/storage-disks-vm
```diff
@@ -170,6 +170,96 @@ if [ "${DEVICE_HOTPLUG}" = "1" ]; then
...
+lxc exec v2 -- test -b /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1rw
+
+# Check the rw driver accepts writes and the ro does not.
+! lxc exec v2 -- dd if=/dev/urandom of=/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro bs=512 count=2 || false
+lxc exec v2 -- dd if=/dev/urandom of=/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1rw bs=512 count=2
+
+echo "==> Checking disk device hotplug support"
+lxc config device add v2 block1 disk source="${testRoot}/allowed1/lxd-block-test" readonly=true
+# shellcheck disable=SC2016
+lxc exec v2 -- sh -c 'for i in $(seq 10); do test -e /sys/block/sdb/ro && break; echo "Waiting for sys file to appear (${i}s)"; sleep 1; done'
```
simondeziel, line 241: Ah, another `sdb` showing up here and a few lines below. Would you mind deriving it from the destination of `/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1`.

Something like `DISK_DEV="$(basename "$(readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1)")"`

tests/storage-disks-vm
```diff
@@ -170,6 +170,96 @@ if [ "${DEVICE_HOTPLUG}" = "1" ]; then
...
+# Hot plug cloud-init:config ISO.
+lxc config device add v2 cloudinit disk source=cloud-init:config
+sleep 3
+BLOCK_SPECIAL_FILE="$(readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_CD-ROM_lxd_cloudinit)"
+lxc exec v2 -- test -b "${BLOCK_SPECIAL_FILE}"
+lxc exec v2 -- mount -t iso9660 -o ro /dev/disk/by-id/scsi-0QEMU_QEMU_CD-ROM_lxd_cloudinit /mnt
+lxc exec v2 -- umount /dev/disk/by-id/scsi-0QEMU_QEMU_CD-ROM_lxd_cloudinit
+lxc config device remove v2 cloudinit
+! lxc exec v2 -- stat "${BLOCK_SPECIAL_FILE}" || false
+! lxc exec v2 -- stat /dev/disk/by-id/scsi-0QEMU_QEMU_CD-ROM_lxd_cloudinit || false
```
simondeziel, line 260: Same comment here: `s/stat/test -e/`
Thanks.

tests/storage-disks-vm
```diff
@@ -170,6 +170,96 @@ if [ "${DEVICE_HOTPLUG}" = "1" ]; then
...
+[ "$(lxc exec v2 -- mount | grep -cF '/srv/ro type')" = "1" ]
+
+# RW disks should use virtio-fs when used with the snap.
+lxc exec v2 -- mount | grep 'lxd_dir1rw on /srv/rw type virtiofs (rw,relatime)'
+
+# RO disks should use virtio-fs when used with the snap but be mounted readonly.
+lxc exec v2 -- mount | grep 'lxd_dir1ro on /srv/ro type virtiofs (ro,relatime)'
+
+# Check UID/GID are correct.
+[ "$(lxc exec v2 -- stat -c '%u:%g' /srv/rw)" = '0:0' ]
```
hamistao, line 202: @simondeziel The tests are failing here because the user and group come out as 65534 (nobody) and not 0 (root). I am unsure how to handle this since changing the device source directory would break the tests due to profile restrictions and neither creating the directory with `sudo` nor changing the tests to expect `65534:65534` seem like appropriate solutions.

tests/storage-disks-vm
```diff
@@ -127,8 +127,8 @@ BLOCK_SPECIAL_FILE="$(lxc exec v2 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU
 lxc exec v2 -- test -b "$BLOCK_SPECIAL_FILE"
 lxc config device remove v2 ${LONG_DEVICE_NAME}
 sleep 3
-! lxc exec v2 -- stat "$BLOCK_SPECIAL_FILE" || false
-! lxc exec v2 -- stat /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_device-----with- || false
+! lxc exec v2 -- test -e "$BLOCK_SPECIAL_FILE" || false
```
tomponline, line 130: This change isn't moving a test and so shouldnt be in a commit called "Move volume related tests to storage-volumes-vm"

tomponline:
@simondeziel @hamistao although I agree we should be checking /dev/disk/by-id most of the time, I think we should be checking at least once that at **boot time** the order the disks are attached is manifesting correctly in the guest such that root drive is /dev/sda, the 2nd disk (by device name) is /dev/sdb etc. This will test that our VM device sorting logic doesn't regress, and that LXD is using it to configure QEMU correctly. As ever consideration of what we are losing when changing tests is important.

https://github.com/canonical/lxd/blob/main/lxd/device/config/devices_sort.go#L58-L75

tests/storage-disks-vm
hamistao: I agree. Please be aware that this is still a draft so there are still some improvements left to be done designwise, I will notify when this is ready for a full review.

hamistao:
@tomponline @simondeziel I suggest merging the `sd*` name handling through #234 as it does not concern the moving of the tests.
Taking this into consideration, this is ready for a review.

PR #234, opened 2024-07-10
Handle `sd*` disk names

hamistao:
#231 is a prerequisite.
This removes the last `sd*` reference and adds tests for `sd*` name assignments with coldplugged disks.

### DIFF
--- a/tests/storage-disks-vm
+++ b/tests/storage-disks-vm
@@ -103,6 +103,11 @@ lxc exec v1 -- rm /srv/rw/lxd-test
 lxc exec v1 -- test -b /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro
 lxc exec v1 -- test -b /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1rw
 
+# Check /dev/sd*
+[ "$(lxc exec v1 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root)" = "/dev/sda" ]
+[ "$(lxc exec v1 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1ro)" = "/dev/sdb" ]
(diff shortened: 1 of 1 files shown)


tomponline:
Ready for rebase

simondeziel (approved):
LGTM. @tomponline you wanted to make sure LXD preserved the disk ordering to match that of the config which I think this PR does.

### Not for that PR but an idea for later

I've been thinking that maybe we could have an additional test where we unplug+plug back a disk to simulate `/dev/sdc` going out and back in. This would verify it takes back its slot so to speak. Maybe that's something we should do with other devices like NICs too. WDYT?

hamistao:
@simondeziel I would say it makes sense to add these checks on this PR, at least the disk related one.

tests/storage-disks-vm
```diff
@@ -112,12 +112,13 @@ lxc config device remove v1 block1rw
 lxc config device remove v1 block1ro
 lxc config device add v1 block1 disk source="${testRoot}/allowed1/lxd-block-test" readonly=true
 # shellcheck disable=SC2016
-lxc exec v1 -- sh -c 'for i in $(seq 10); do test -e /sys/block/sdb/ro && break; echo "Waiting for sys file to appear (${i}s)"; sleep 1; done'
-[ "$(lxc exec v1 -- cat /sys/block/sdb/ro)" -eq 1 ]
+DISK_DEV="$(basename "$(lxc exec v1 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_block1)")"
+lxc exec v1 -- sh -c "for i in \$(seq 10); do test -e /sys/block/${DISK_DEV}/ro && break; echo \"Waiting for sys file to appear (\${i}s)\"; sleep 1; done"
```
tomponline, line 121: Why the escape character changes?

tomponline:
> This would verify it takes back its slot so to speak

Sounds good!

tests/storage-disks-vm
simondeziel: Probably needed to have `${DISK_DEV}` expanded but not the other `$`.

tomponline (approved):

tests/storage-disks-vm
hamistao: That is correct

PR #238, opened 2024-07-11
Update dir hotplug tests

hamistao:
Hotplugging directories has been allowed since canonical/lxd@fff4b19. The test was passing anyway because the command syntax was incorrect, missing the mount path argument.


tomponline (approved):
Good spot!

PR #240, opened 2024-07-12
Test image export over devlxd socket

markylaing:
Adds tests for the devlxd socket in a container. These tests are based on the devlxd tests for VMs, with additional testing for image export over the socket.

Note: I wasn't sure where else to add this test other than in the github workflow.

Also note: This test will fail on edge until https://github.com/canonical/lxd/pull/13730 is merged.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,6 +85,7 @@ jobs:
           - container-copy
           - cpu-vm
           - devlxd-vm
+          - devlxd-container
           - docker
           - efi-vars-editor-vm
           - interception


tests/devlxd-container
```diff
@@ -0,0 +1,121 @@
...
+  lxc exec c1 -- curl -s --unix-socket /dev/lxd/sock -X PATCH -d '{"state":"Ready"}' http://custom.socket/1.0
+  [ "$(lxc config get c1 volatile.last_state.ready)" = "true" ]
+  lxc stop -f c1
+  sleep 5
+  [ "$(lxc config get c1 volatile.last_state.ready)" = "false" ]
+
+  lxc start c1
+
+  # We need snapd to be ready before the next tests. Finagle the waitSnapdSeed function definition into the container and run it.
+  lxc exec c1 -- sh -c "$(type waitSnapdSeed | tail -n +2); waitSnapdSeed"
```
simondeziel, line 84: I need to better understand that trick as it looks super useful! Thanks!

simondeziel (commented):
What would you think of `git mv tests/devlxd-vm tests/devlxd` and then put a `for i in c1 v1; do` loop in it?

tomponline:
Is this ready ?

markylaing:
> What would you think of `git mv tests/devlxd-vm tests/devlxd` and then put a `for i in c1 v1; do` loop in it?

Sorry @simondeziel I don't understand what you mean here. Is this so that we run the same devlxd tests for both containers and VMs? If so, the image export will fail for VMs as that's not implemented yet.

markylaing:
> Is this ready ?

I'll rerun the test on latest/edge to make sure it works and then I think so pending Simon's review.

markylaing:
The test is passing on latest/edge: https://github.com/canonical/lxd-ci/actions/runs/9908022697/job/27557147131

So I think we're good to go :)

tomponline:
@markylaing its failing on 4.0/edge, which wont get the fixes from latest/edge, so think we need to add conditional logic in the test so that the devlxd test passes for 4.0/edge (i.e only test the relevant parts).

simondeziel:
> > What would you think of `git mv tests/devlxd-vm tests/devlxd` and then put a `for i in c1 v1; do` loop in it?
> 
> Sorry @simondeziel I don't understand what you mean here. Is this so that we run the same devlxd tests for both containers and VMs?

Yeah, sorry for being particularly unclear but you got the intention right ;)

> If so, the image export will fail for VMs as that's not implemented yet.

I was thinking that very little would be different and it would be easy to skip the bits that are known to be missing, thanks to API detection.

The goal would be to avoid creating yet another parallel job (albeit a quick one) and duplicate a test script almost entirely.

markylaing:
> @markylaing its failing on 4.0/edge, which wont get the fixes from latest/edge, so think we need to add conditional logic in the test so that the devlxd test passes for 4.0/edge (i.e only test the relevant parts).

Ok will do. I'm not sure about 5.0/edge though, is that still receiving bugfixes? That test failed to query the `ubuntu` remote, but I think it ought to pass if we backport https://github.com/canonical/lxd/pull/13730/commits/24111f6af1946a8e8936d33b3e8df3d0f5c865cd

4.0 is failing before it gets to the image export part, so I can figure out which ones are failing and skip those, or skip the test entirely. What do you think?

tomponline:
> Ok will do. I'm not sure about 5.0/edge though, is that still receiving bugfixes? That test failed to query the `ubuntu` remote, but I think it ought to pass if we backport [canonical/lxd@24111f6](https://github.com/canonical/lxd/commit/24111f6af1946a8e8936d33b3e8df3d0f5c865cd)

We'll do one more point release so if you can open a PR to backport that change great, thanks

tomponline:
> 4.0 is failing before it gets to the image export part, so I can figure out which ones are failing and skip those, or skip the test entirely. What do you think?

I would say skip the ones that dont work if possible.

tomponline:
@markylaing you can use 4.0/stable as a guide as to what "works" today

tomponline:
@markylaing ive moved this back to draft, let me know when ready for review again ta.

markylaing:
@tomponline passing on 4.0/edge, 5.0/edge, 5.21/edge, and latest/edge.

tests/devlxd-container
```diff
@@ -0,0 +1,134 @@
...
+lxc profile device add default eth0 nic network=lxdbr0
+
+poolName="ctpool$$"
+poolDriver=dir
+
+echo "==> Create storage pool using driver ${poolDriver}"
+lxc storage create "${poolName}" "${poolDriver}"
+
+echo "==> Create container and boot"
+lxc launch "${TEST_IMG:-ubuntu-daily:22.04}" c1 -s "${poolName}" -c security.nesting=true
```
tomponline, line 21: should be 24.04 right?

tests/devlxd-container
```diff
@@ -0,0 +1,134 @@
...
+# Below tests for image export rely on an event being emitted on export. This will not be backported to 4.0 or 5.0,
+# so exit the test here.
+if [ "${LXD_SNAP_CHANNEL}" = "4.0/edge" ] || [ "${LXD_SNAP_CHANNEL}" = "5.0/edge" ]; then
+  echo "Skipping devlxd image export tests, not supported"
+  FAIL=0
+  exit
+fi
+
+# We need snapd to be ready before the next tests. Finagle the waitSnapdSeed function definition into the container and run it.
+lxc exec c1 -- sh -c "$(type waitSnapdSeed | tail -n +2); waitSnapdSeed"
```
tomponline, line 101: did "declare" work btw?

RE https://chat.canonical.com/canonical/pl/851h6r1a4tnutg39ddysfywxpw

tests/devlxd-container
```diff
@@ -0,0 +1,134 @@
...
+lxc exec c1 -- snap install lxd --channel="${LXD_SNAP_CHANNEL}"
+lxc exec c1 -- /snap/bin/lxd init --auto
+
+if hasNeededAPIExtension devlxd_images && hasNeededAPIExtension event_lifecycle; then
+  lxc monitor --type lifecycle --format json > monitor.json 2>&1 &
+  echo $! > monitor.pid
+fi
+
+# Launch an image we know is on the host.
+lxc exec c1 -- /snap/bin/lxc launch "${TEST_IMG:-ubuntu-daily:22.04}" c1c1
```
tomponline, line 112: 24.04 right?

tests/devlxd-container
```diff
@@ -0,0 +1,134 @@
...
+
+# We need snapd to be ready before the next tests. Finagle the waitSnapdSeed function definition into the container and run it.
+lxc exec c1 -- sh -c "$(type waitSnapdSeed | tail -n +2); waitSnapdSeed"
+lxc exec c1 -- snap remove --purge lxd || true
+lxc exec c1 -- snap install lxd --channel="${LXD_SNAP_CHANNEL}"
+lxc exec c1 -- /snap/bin/lxd init --auto
+
+if hasNeededAPIExtension devlxd_images && hasNeededAPIExtension event_lifecycle; then
+  lxc monitor --type lifecycle --format json > monitor.json 2>&1 &
+  echo $! > monitor.pid
```
tomponline, line 108: Can you just store this as a variable rather than a file, less to clear up or be left behind.

tomponline (commented):
Couple of nits.

Also, can this work with ubuntu-minimal-daily or is there no benefit because of the extras needed to be installed?

tests/devlxd-container
markylaing: It works!

markylaing:
> Couple of nits.
> 
> Also, can this work with ubuntu-minimal-daily or is there no benefit because of the extras needed to be installed?

Updated to use ubuntu-minimal-daily as well as the other nits.

tomponline:
shell check is failing

tests/devlxd-container
tomponline: cool, its a bit easier to understand than type!

markylaing:
> > Couple of nits.
> > Also, can this work with ubuntu-minimal-daily or is there no benefit because of the extras needed to be installed?
> 
> Updated to use ubuntu-minimal-daily as well as the other nits.

Just need to fix the new shellcheck issues :grimacing:

markylaing:
@tomponline @simondeziel the code test is failing because `declare` is undefined for POSIX shells. We'll need to specify `--shell=bash` to get this to pass. Or I can revert to using `type {function-name} | tail -n+2`

tomponline:
> We'll need to specify `--shell=bash` to get this to pass. Or I can revert to using `type {function-name} | tail -n+2`

shell bash is fine, with a comment explaining why please.

markylaing:
> > We'll need to specify `--shell=bash` to get this to pass. Or I can revert to using `type {function-name} | tail -n+2`
> 
> shell bash is fine, with a comment explaining why please.

It doesn't seem to be possible to specify `--shell` with the action (https://github.com/redhat-plumbers-in-action/differential-shellcheck?tab=readme-ov-file#configuration-options). Instead I've just ignored the error with a shellcheck disable comment (and explained why).

tomponline:
> > > We'll need to specify `--shell=bash` to get this to pass. Or I can revert to using `type {function-name} | tail -n+2`
> > 
> > 
> > shell bash is fine, with a comment explaining why please.
> 
> It doesn't seem to be possible to specify `--shell` with the action (https://github.com/redhat-plumbers-in-action/differential-shellcheck?tab=readme-ov-file#configuration-options). Instead I've just ignored the error with a shellcheck disable comment (and explained why).

sounds good

markylaing:
All tests passing and issues addressed.

tomponline (approved):

PR #245, opened 2024-07-18
Storage operation integrity checks

simondeziel:
This is just the first cut at storage operation integrity checks. I intend to implement your `lxc file mount` suggestion for containers in a future PR but if you prefer it to be bundled here, that's OK too.

### DIFF
--- a/tests/storage-vm
+++ b/tests/storage-vm
@@ -349,6 +349,10 @@ for poolDriver in $poolDriverList; do
                 echo "==> Checking VM root disk size is 8GiB"
                 [ "$(($(lxc exec v1 -- blockdev --getsize64 /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root) / GiB))" -eq "8" ]
         fi
+        V1_ROOT_GPT_DISK_ID="$(lxc exec v1 -- fdisk --list /dev/sda | awk '/^Disk identifier:/ {print $3}')"
+        V1_EFI_PART="$(lxc exec v1 -- fdisk --list /dev/sda | awk '/EFI System$/ {print $1}')"
+        lxc exec v1 -- umount "${V1_EFI_PART}"
+        V1_EFI_SHA256SUM="$(lxc exec v1 -- sha256sum "${V1_EFI_PART}")"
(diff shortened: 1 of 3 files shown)


tests/storage-vm
```diff
@@ -369,6 +373,21 @@ for poolDriver in $poolDriverList; do
...
+        lxc exec v2 -- umount "${V2_EFI_PART}"
+        V2_EFI_SHA256SUM="$(lxc exec v2 -- sha256sum "${V2_EFI_PART}")"
+        [ "${V1_EFI_SHA256SUM}" = "${V2_EFI_SHA256SUM}" ]
+
+        echo "==> Check the GPT disk ID of the root disk did not change after the copy"
+        V2_ROOT_GPT_DISK_ID="$(lxc exec v2 -- fdisk --list /dev/sda | awk '/^Disk identifier:/ {print $3}')"
+        [ "${V1_ROOT_GPT_DISK_ID}" = "${V2_ROOT_GPT_DISK_ID}" ]
+
+        echo "==> Check the GPT backup table was copied at the new end of the disk"
+        # `fdisk --list` would print `The backup GPT table is corrupt, but the primary appears OK, so that will be used.` to stderr
```
tomponline, line 386: is this a bad sign? If not a comment here explaining why would be useful.

tests/storage-vm
```diff
@@ -185,7 +185,7 @@ for poolDriver in $poolDriverList; do
 
         echo "==> Checking VM can be refreshed remotely (different storage pool)"
         lxc snapshot v1
-        lxc copy v1 localhost:v2 --refresh
+        lxc copy v1 localhost:v2 --refresh -s "${poolName}2"
```
tomponline, line 188: Good catch

tomponline (approved):
Thanks. This is a great first addition for these checks.

One area where I think bit-for-bit comparison for VM copies is especially important is the refresh stage using snapshots for optimized syncing, as different storage drivers use different methods, and ceph in particular is somewhat complex (and thus potential for error increases) and ceph's mechanism doesn't detect corruption and so you can end up with an invalid disk if not done correctly.

PR #248, opened 2024-07-26
tests/snapd: Add quick check for snap preseed handling

boltmark:
This PR adds a quick check to ensure that preseeds set via snap are handled, and makes updates to `tests/snapd` to remove redundant testing. 

`tests/snapd` now runs in the pipeline.

### DIFF
--- a/tests/snapd
+++ b/tests/snapd
@@ -82,8 +82,6 @@ lxc profile device add default root disk pool=default path=/
 lxc network create lxdbr0
 lxc profile device add default eth0 nic nictype=bridged parent=lxdbr0 name=eth0
 
-snap set lxd ceph.builtin=true
-
 systemctl restart snap.lxd.daemon.service
 
 for _ in $(seq 60); do
@@ -94,50 +92,5 @@ done
 useradd -m ${TEST_USER} -G lxd
 LXC=$(command -v lxc)
 su ${TEST_USER} -c "${LXC} info"
-
-# Container creation test
-lxc launch images:alpine/edge unprivileged
-lxc launch images:alpine/edge privileged -c security.privileged=true
-lxc launch images:alpine/edge isolated -c security.idmap.isolated=true
-
-# Storage backends test
-if modprobe -q -n btrfs; then
-    lxc storage create test-btrfs btrfs
-    lxc launch images:alpine/edge -s test-btrfs btrfs
-fi
-
-if modprobe -q -n zfs && ! grep -q ^0.7 < /sys/module/zfs/version; then
-    lxc storage create test-zfs zfs
-    lxc launch images:alpine/edge -s test-zfs zfs
-fi
-
-if [ "${ID}" != "ubuntu-core" ] && [ "${ID}" != "elementary" ] && modprobe -q -n dm_mod; then
-    lxc storage create test-lvm lvm volume.size=25MB
-    lxc launch images:alpine/edge -s test-lvm lvm
-fi
-
-if [ "${ID}" != "ubuntu-core" ] && [ "${VERSION_CODENAME:-""}" != "xenial" ] && [ "${ID}" != "centos" ] && modprobe -q -n rbd; then
-    curl http://canonical-lxd.stgraber.org/config/ceph-snap.sh | sh -eu
-
-    lxc storage create test-ceph ceph "ceph.osd.pool_name=$(mktemp -u snaptest-XXX)" volume.size=25MB ceph.osd.pg_num=1
-    lxc launch images:alpine/edge -s test-ceph ceph
-fi
-
-if [ "${track}" != "3.0" ]; then
-    lxc query /1.0/resources
-fi
-
-lxc file push /usr/bin/snap unprivileged/tmp/
-echo foo | lxc file push - unprivileged/tmp/foo
-[ "$(lxc file pull unprivileged/tmp/foo -)" = "foo" ]
-
-# Basic validation of running containers
-for container in $(lxc list --fast | tail -n+3 | grep "^| " | cut -d' ' -f2); do
-    lxc exec "${container}" -- ls /
-    lxc info "${container}"
-done
-
-sleep 10
-lxc list
 
 RET=0
--- a/tests/snapd
+++ b/tests/snapd
@@ -3,26 +3,10 @@ set -eux
 
 TEST_USER="testuser"
 
-# Configure to use the proxy
-curl -s http://canonical-lxd.stgraber.org/config/snapd.sh | sh
+# Install lxd without starting daemon
+install_lxd false
 
-if [ "${track}" = "latest" ]; then
-    snapd_channel=${channel}
-else
-    snapd_channel=${track}/${channel}
-fi
-
-echo "==> Installing LXD snap from ${track}/${channel}"
-snap remove lxd || true
-snap version
-snap list
-snap install lxd --channel="${snapd_channel}" --cohort=+
-
-for _ in $(seq 60); do
-    lxd waitready --timeout=10 >/dev/null 2>&1 && break
-done
-
-lxc --version
+lxd waitready --timeout=300
 
 # LXD setup
 lxc storage create default dir
@@ -30,12 +14,6 @@ lxc profile device add default root disk pool=default path=/
 
 lxc network create lxdbr0
 lxc profile device add default eth0 nic nictype=bridged parent=lxdbr0 name=eth0
-
-systemctl restart snap.lxd.daemon.service
-
-for _ in $(seq 60); do
-    lxd waitready --timeout=10 >/dev/null 2>&1 && break
-done
 
 # Check that user access works
 useradd -m ${TEST_USER} -G lxd
--- a/tests/snapd
+++ b/tests/snapd
@@ -1,29 +1,24 @@
 #!/bin/sh
 set -eux
 
-# Source host information
-# shellcheck disable=SC1091
-[ -e "/etc/os-release" ] && . /etc/os-release
-
-track=${1:-latest}
-channel=${2:-stable}
-
 export PATH="${PATH}:/var/lib/snapd/snap/bin:/snap/bin"
 
 RET=1
+TEST_USER="testuser"
+
 cleanup() {
     REBOOT=0
 
     set +e
     echo "==> Cleaning up"
-    for container in $(lxc list --fast | tail -n+3 | grep "^| " | cut -d' ' -f2); do
+        for container in $(lxc list --fast | tail -n+3 | grep "^| " | cut -d' ' -f2); do
         lxc delete "${container}" -f || true
     done
 
     for image in $(lxc image list | tail -n+3 | grep "^| " | cut -d'|' -f3 | sed "s/^ //g"); do
         lxc image delete "${image}" || true
     done
-
+    
     for profile in $(lxc profile list | tail -n+3 | grep "^| " | cut -d' ' -f2); do
         printf 'config: {}\ndevices: {}' | lxc profile edit "${profile}" || true
     done
@@ -96,14 +91,9 @@ for _ in $(seq 60); do
 done
 
 # Check that user access works
-if [ "${ID}" != "ubuntu-core" ]; then
-    (
-        set -e
-        usermod -G lxd -a jenkins || true
-        LXC=$(which lxc)
-        su jenkins -c "${LXC} info"
-    )
-fi
+useradd -m ${TEST_USER} -G lxd
+LXC=$(command -v lxc)
+su ${TEST_USER} -c "${LXC} info"
 
 # Container creation test
 lxc launch images:alpine/edge unprivileged
--- a/tests/snapd
+++ b/tests/snapd
@@ -6,7 +6,13 @@ TEST_USER="testuser"
 # Install lxd without starting daemon
 install_lxd false
 
+# Set preseed via snap
+snap set lxd daemon.preseed='{"config": {"core.https_address": "[::]:8443"}}'
+
 lxd waitready --timeout=300
+
+# Check that preseed via snap works
+lxc config show | grep -qF "core.https_address: '[::]:8443'" 
 
 # LXD setup
 lxc storage create default dir
--- a/tests/snapd
+++ b/tests/snapd
@@ -1,58 +1,7 @@
 #!/bin/sh
 set -eux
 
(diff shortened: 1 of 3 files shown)


tomponline (approved):
Excellent thanks

tomponline:
Oh looks like the snapd test doesnt run, is that on purpose @simondeziel ?

tomponline:
@boltmark please can you enable this test in this section of the workflow and we'll see if it passes:

https://github.com/canonical/lxd-ci/blob/main/.github/workflows/tests.yml#L84-L177

boltmark:
> @boltmark please can you enable this test in this section of the workflow and we'll see if it passes:
> 
> https://github.com/canonical/lxd-ci/blob/main/.github/workflows/tests.yml#L84-L177

Looks like `su jenkins -c '/usr/sbin/lxc info'` is failing with error message
`su: user jenkins does not exist or the user entry does not contain all the required fields`

tomponline:
> > @boltmark please can you enable this test in this section of the workflow and we'll see if it passes:
> > https://github.com/canonical/lxd-ci/blob/main/.github/workflows/tests.yml#L84-L177
> 
> Looks like `su jenkins -c '/usr/sbin/lxc info'` is failing with error message `su: user jenkins does not exist or the user entry does not contain all the required fields`

Please can you update to remove Jenkins references

tomponline:
You may need to update to use helpers and vars the other tests use

boltmark:
> You may need to update to use helpers and vars the other tests use

If I am understanding correctly, we have the below check just to ensure we can use lxd as a user in the `lxd` group. Is there a particular reason we are using the `jenkins` user? If not, I've updated this check to just create and use a new user, but please let me know if I'm missing something here.

```bash
if [ "${ID}" != "ubuntu-core" ]; then
    (
        set -e
        usermod -G lxd -a jenkins || true
        LXC=$(which lxc)
        su jenkins -c "${LXC} info"
    )
    )
fi
```

tomponline:
I think we can remove references to ubuntu core and jenkins.

tests/snapd
```diff
@@ -59,20 +58,12 @@ cleanup() {
...
-    snapd_channel=${track}/${channel}
-fi
-
-echo "==> Installing LXD snap from ${track}/${channel}"
+echo "==> Installing LXD snap from ${LXD_SNAP_CHANNEL}"
 snap remove lxd || true
 snap version
 snap list
-snap install lxd --channel="${snapd_channel}" --cohort=+
+snap install lxd --channel="${LXD_SNAP_CHANNEL}" --cohort=+
```
simondeziel, line 65: There is the `install_lxd` helper that should replace most of those.

tests/snapd
```diff
@@ -59,20 +58,12 @@ cleanup() {
...
+snap set lxd daemon.preseed='{"config": {"core.https_address": "[::]:8443"}}'
```
simondeziel, line 66: If the intention is to have a network listener, `lxd init --auto --network-address "[::]" --network-port "8443"` should be more straight forward.

tests/snapd
```diff
@@ -99,12 +90,19 @@ done
...
+        usermod -G lxd -a ${test_user}
         LXC=$(which lxc)
-        su jenkins -c "${LXC} info"
+        su ${test_user} -c "${LXC} info"
+        userdel -r ${test_user} 2>/dev/null || true
     )
 fi
 
+# Check that preseed via snap works
+lxc config show | grep -qF "core.https_address: '[::]:8443'"
```
simondeziel, line 104: Ah, I now understand the earlier `snap set lxd daemon.preseed`. Would you mind putting the comment+check closer to the `snap set`?

tests/snapd
```diff
@@ -5,8 +5,7 @@ set -eux
 # shellcheck disable=SC1091
 [ -e "/etc/os-release" ] && . /etc/os-release
 
-track=${1:-latest}
-channel=${2:-stable}
+track="${LXD_SNAP_CHANNEL%%/*}"
```
simondeziel, line 8: AFAICT, we no longer need the track as we don't care for LXD 3.0 as it's EOL.

tests/snapd
```diff
@@ -59,20 +58,12 @@ cleanup() {
...
+echo "==> Installing LXD snap from ${LXD_SNAP_CHANNEL}"
 snap remove lxd || true
 snap version
 snap list
-snap install lxd --channel="${snapd_channel}" --cohort=+
+snap install lxd --channel="${LXD_SNAP_CHANNEL}" --cohort=+
+snap set lxd daemon.preseed='{"config": {"core.https_address": "[::]:8443"}}'
 
 for _ in $(seq 60); do
     lxd waitready --timeout=10 >/dev/null 2>&1 && break
```
simondeziel, line 69: This should be a single `lxd waitready --timeout=300` like we do everywhere else.

tests/snapd
```diff
@@ -99,12 +90,19 @@ done
 if [ "${ID}" != "ubuntu-core" ]; then
     (
         set -e
-        usermod -G lxd -a jenkins || true
+
+        test_user="testuser"
+        useradd -m ${test_user} || true
+        usermod -G lxd -a ${test_user}
```
simondeziel, line 96: `useradd` supports adding additional groups via the `--groups` option.

tests/snapd
```diff
@@ -99,12 +90,19 @@ done
...
         LXC=$(which lxc)
```
simondeziel, line 97: Please replace `which lxc` by `command -v lxc`.

tests/snapd
```diff
@@ -99,12 +90,19 @@ done
...
+        userdel -r ${test_user} 2>/dev/null || true
     )
 fi
 
+# Check that preseed via snap works
+lxc config show | grep -qF "core.https_address: '[::]:8443'" 
+
 # Container creation test
 lxc launch images:alpine/edge unprivileged
 lxc launch images:alpine/edge privileged -c security.privileged=true
```
simondeziel, line 108: Let's use `ubuntu-minimal-daily:24.04` here please.

simondeziel (changes requested):
I'm not entirely sure about the `ceph.builtin=true` usage in this script. Seems like a leftover from when it was running on the Jenkins CI.

simondeziel:
> Oh looks like the snapd test doesnt run, is that on purpose @simondeziel ?

Indeed, it was never "ported" to GitHub CI.

tests/snapd
boltmark: Did see this helper, but in this case we need to preseed before starting the daemon, so we can't use `install_lxd` because it runs `waitready` after installation.

tests/snapd
boltmark: Did this to handle the case where the user already exists, but I suppose this should never happen. Updated to use -G with `useradd`.

tests/snapd
boltmark: Updated all uses of `alpine/edge` to `ubuntu-minimal-daily:24.04` since it seems like this is what is used in other tests.

boltmark:
@simondeziel I appreciate the review! I've made updates and left a few comments.

In the pipeline this test seems to fail to connect to `images.lxd.canonical.com`. This does not happen for me locally. Any idea why this is failing here?

```
$ lxc launch images:ubuntu-minimal-daily:24.04 unprivileged
Launching unprivileged
Error: Failed instance creation: Failed getting remote image info: Failed getting image: Failed parsing stream: Get "https://images.lxd.canonical.com/streams/v1/index.json": Unable to connect to: images.lxd.canonical.com:443 ([dial tcp 127.0.0.1:443: connect: connection refused])
```

tomponline:
> @simondeziel I appreciate the review! I've made updates and left a few comments.
> 
> In the pipeline this test seems to fail to connect to `images.lxd.canonical.com`. This does not happen for me locally. Any idea why this is failing here?
> 
> ```
> $ lxc launch images:ubuntu-minimal-daily:24.04 unprivileged
> Launching unprivileged
> Error: Failed instance creation: Failed getting remote image info: Failed getting image: Failed parsing stream: Get "https://images.lxd.canonical.com/streams/v1/index.json": Unable to connect to: images.lxd.canonical.com:443 ([dial tcp 127.0.0.1:443: connect: connection refuse

No need for " images:" prefix

boltmark:
> > @simondeziel I appreciate the review! I've made updates and left a few comments.
> > In the pipeline this test seems to fail to connect to `images.lxd.canonical.com`. This does not happen for me locally. Any idea why this is failing here?
> > ```
> > $ lxc launch images:ubuntu-minimal-daily:24.04 unprivileged
> > Launching unprivileged
> > Error: Failed instance creation: Failed getting remote image info: Failed getting image: Failed parsing stream: Get "https://images.lxd.canonical.com/streams/v1/index.json": Unable to connect to: images.lxd.canonical.com:443 ([dial tcp 127.0.0.1:443: connect: connection refuse
> > ```
> 
> No need for " images:" prefix

Oops. Thanks for catching that @tomponline.

tomponline:
Looks like the tests are failing for snapd test

tomponline:
@boltmark we are interested in latest/edge tests passing for this, as ive not backported your fix to 5.21/edge yet.

The error you're seeing is:

> Error: Failed instance creation: Failed creating instance from image: Unpack failed: Failed to run: unsquashfs -f -d /var/snap/lxd/common/lxd/storage-pools/test-lvm/images/cf50a6b79620d2cfe6337c389a6a70f78dd1a6d092583786380a0bcd86cf51e2/rootfs -n /var/snap/lxd/common/lxd/images/cf50a6b79620d2cfe6337c389a6a70f78dd1a6d092583786380a0bcd86cf51e2.rootfs: exit status 1 (Write on output file failed because No space left on device
Retrieving image: Unpacking image: 100% (761.55MB/s)
FATAL ERROR: writer: failed to write file /var/snap/lxd/common/lxd/storage-pools/test-lvm/images/cf50a6b79620d2cfe6337c389a6a70f78dd1a6d092583786380a0bcd86cf51e2/rootfs/usr/bin/dpkg-divert)


https://github.com/canonical/lxd-ci/actions/runs/10205740488/job/28237210776?pr=248

boltmark:
@tomponline I've removed the ceph check, and the snapd test now only fails on 5.21 (expected as preseed fix not yet backported), and 4.0/edge (only on 22.04) due to `Error: Required tool 'zpool' is missing` when creating a zfs storage pool.

Is this related to https://discuss.linuxcontainers.org/t/ubuntu-22-04-lxd-5-0-2-required-tool-zpool-is-missing-after-apt-upgrade/18791/14? If it is, is there a similar fix for 4.0 or should we skip this check for 4.0 on 22.04 in the testing?

boltmark:
> Let's get rid of the zfs check too for now, this is tested elsewhere

Sounds good. Should we just get rid of all storage tests (only btrfs and lvm remaining) in the `snapd` test? I see these are tested more thoroughly in `storage-vm`.

tomponline:
Yeah good idea

tests/snapd
```diff
@@ -87,55 +79,22 @@ lxc profile device add default root disk pool=default path=/
...
 # Check that user access works
-if [ "${ID}" != "ubuntu-core" ]; then
-    (
-        set -e
-        usermod -G lxd -a jenkins || true
-        LXC=$(which lxc)
-        su jenkins -c "${LXC} info"
-    )
-fi
+set -e
```
simondeziel, line 87: This `set -e` is probably not wanted as it would prevent catching any error but that of the last command in the script.

tests/snapd
```diff
@@ -87,55 +79,22 @@ lxc profile device add default root disk pool=default path=/
...
-    lxc storage create test-ceph ceph "ceph.osd.pool_name=$(mktemp -u snaptest-XXX)" volume.size=25MB ceph.osd.pg_num=1
-    lxc launch images:alpine/edge -s test-ceph ceph
-fi
-
-if [ "${track}" != "3.0" ]; then
-    lxc query /1.0/resources
-fi
+lxc launch "${IMAGE}" unprivileged
+lxc launch "${IMAGE}" privileged -c security.privileged=true
+lxc launch "${IMAGE}" isolated -c security.idmap.isolated=true
```
simondeziel, line 97: I think those are potentially unneeded now that we have [`tests/container`](https://github.com/canonical/lxd-ci/blob/main/tests/container)

tests/snapd
boltmark: Removed

tests/snapd
boltmark: Removed container-related checks

tests/snapd
```diff
@@ -5,9 +5,6 @@ set -eux
 # shellcheck disable=SC1091
 [ -e "/etc/os-release" ] && . /etc/os-release
```
simondeziel, line 6: This can go as it was used for `[ "${ID}" != "ubuntu-core" ]` which is no longer there.

tests/snapd
```diff
@@ -36,6 +25,8 @@ cleanup() {
         lxc storage delete "${storage}" || REBOOT=1
     done
 
+    userdel -r ${test_user} 2>/dev/null || true
```
simondeziel, line 28: Could you drop that `cleanup()` function and instead do something similar to https://github.com/canonical/lxd-ci/blob/main/tests/efi-vars-editor-vm#L76-L86 while keeping the `userdel` part, ofc.

There is already a `cleanup()` function in `bin/helpers` so this local one is shadowing it anyway.

tests/snapd
```diff
@@ -87,65 +71,15 @@ lxc profile device add default root disk pool=default path=/
...
-    lxc exec "${container}" -- ls /
-    lxc info "${container}"
-done
+test_user="testuser"
+useradd -m ${test_user} -G lxd
+LXC=$(command -v lxc)
+su ${test_user} -c "${LXC} info"
 
 sleep 10
 lxc list
```
simondeziel, line 85: This sleep and list now seems superfluous now that no instances are launched.

tests/snapd
simondeziel: Could you update `install_lxd()` to accept an argument on whether it should run `waitready` or not?
Also, adding `--cohort=+` to the `install_lxd()` would make a lot of sense.

I'm coming back to this due to the `snap remove lxd || true` lacking the `--purge` :)

boltmark: Updated

tests/snapd
boltmark: Done, thanks for the reference link!

simondeziel (approved):
Thanks!

PR #251, opened 2024-08-01
VM live migration test fix

markylaing:
This PR gets the VM live migration test passing by waiting for the instance to boot before performing the migration.


markylaing:
@tomponline this gets the initial test passing. I'll work on a separate PR to format the volume and ensure it's working after migration.

tomponline (approved):
Ta

PR #253, opened 2024-08-02
Fix intermittent failure on snapd seed

markylaing:
The `devlxd-container` tests are timing out when waiting for snapd to be seeded inside the container that will run a nested LXD.

This PR calls `waitInstanceBooted` after the container starts, then waits indefinitely for snap to be seeded. This approach is that same as in the `cluster` test.

### DIFF
--- a/bin/helpers
+++ b/bin/helpers
@@ -1,12 +1,14 @@
 # shellcheck disable=SC2148
 
 # waitSnapdSeed: wait for snapd to be seeded.
+# Optional argument: timeout in seconds, defaults to 60.
 waitSnapdSeed() (
+  waitSecs="${1:-60}"
   set +x
-  if timeout 60 snap wait system seed.loaded; then
+  if timeout "${waitSecs}" snap wait system seed.loaded; then
     return 0 # Success.
   fi
-  echo "snapd not seeded after 60s"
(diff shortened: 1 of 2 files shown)


markylaing:
@tomponline well increasing the timeout doesn't fix it unfortunately: https://github.com/canonical/lxd-ci/actions/runs/10218115715/job/28273490182?pr=253#step:8:496

Do you know of any reason why snapd will take so long to seed? This is inside an `ubuntu-minimal-daily:24.04` container with `security.nesting=true`.

tomponline:
> Do you know of any reason why snapd will take so long to seed? This is inside an `ubuntu-minimal-daily:24.04` container with `security.nesting=true`.

Are there any other tests where we wait for snapd to seed, can you look what is the difference here, perhaps it is nesting or the presence of /dev/lxd?

markylaing:
> Are there any other tests where we wait for snapd to seed, can you look what is the difference here, perhaps it is nesting or the presence of /dev/lxd?

It looks like in `tests/cluster` we wait for the snapd seed indefinitely after calling `waitInstanceBooted`. I've just updated the test follow the same pattern.

markylaing:
@tomponline I think this is good to go now.

tomponline (approved):

PR #255, opened 2024-08-06
Test devlxd image export for virtual machines

markylaing:
Amends `devlxd-vm` to test for image export over devlxd. To test this locally, I added functionality to the tests to allow sideloading the lxd-agent.

### DIFF
--- a/tests/devlxd-vm
+++ b/tests/devlxd-vm
@@ -11,7 +11,7 @@ install_lxd
 lxc network create lxdbr0
 lxc profile device add default eth0 nic network=lxdbr0
 
-IMAGE="${TEST_IMG:-ubuntu-daily:24.04}"
+IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
 
 poolName="vmpool$$"
 poolDriver=dir
@@ -20,7 +20,12 @@ echo "==> Create storage pool using driver ${poolDriver}"
 lxc storage create "${poolName}" "${poolDriver}"
 
 echo "==> Create VM and boot"
-lxc launch "${IMAGE}" v1 --vm -s "${poolName}"
+if hasNeededAPIExtension devlxd_images_vm; then
+  lxc launch "${IMAGE}" v1 --vm -s "${poolName}" -c security.devlxd.images=true
+else
+  lxc launch "${IMAGE}" v1 --vm -s "${poolName}"
+fi
+
 waitInstanceReady v1
 lxc info v1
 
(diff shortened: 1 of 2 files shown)


markylaing:
This needs some adjustments to pass on `{4.0,5.0}/edge` as the `security.devlxd.images` config key is not available for VMs in those versions.

markylaing:
> This needs some adjustments to pass on `{4.0,5.0}/edge` as the `security.devlxd.images` config key is not available for VMs in those versions.

@tomponline do you think it's worth adding an API extension for this? E.g. `devlxd_images_vm`

tomponline:
> @tomponline do you think it's worth adding an API extension for this? E.g. `devlxd_images_vm`

Yeah i think so

tests/devlxd-vm
```diff
@@ -20,7 +20,12 @@ echo "==> Create storage pool using driver ${poolDriver}"
 lxc storage create "${poolName}" "${poolDriver}"
 
 echo "==> Create VM and boot"
-lxc launch "${IMAGE}" v1 --vm -s "${poolName}"
+if hasNeededAPIExtension devlxd_images; then
```
simondeziel, line 23: `devlxd_images` is in `4.0/stable`:

```
root@f1:~# lxc --version
4.0.10
root@f1:~# lxc info | grep devlxd_images
- devlxd_images
```

To be amended with `devlxd_images_vm` when it's a thing?

tests/devlxd-vm
```diff
@@ -90,14 +95,47 @@ fi
...
+    lxc exec v1 -- systemctl restart snap.lxd.daemon
+fi
+if [ -n "${LXD_AGENT_SIDELOAD_PATH:-}" ]; then
+  lxc file push "${LXD_AGENT_SIDELOAD_PATH}" "v1/root/$(basename "${LXD_AGENT_SIDELOAD_PATH}")"
+  lxc exec v1 -- mount --bind "$(basename "${LXD_AGENT_SIDELOAD_PATH}")" /snap/lxd/current/bin/lxd-agent
+  lxc exec v1 -- systemctl restart snap.lxd.daemon
+fi
+
+monitorPID=""
+if hasNeededAPIExtension devlxd_images && hasNeededAPIExtension event_lifecycle; then
```
simondeziel, line 121: LXD 4.0 supports both `devlxd_images` and `event_lifecycle` so maybe we could drop the conditional altogether?

tests/devlxd-vm
```diff
@@ -90,14 +95,47 @@ fi
...
+  lxc monitor --type lifecycle --format json > monitor.json 2>&1 &
+  monitorPID="${!}"
+fi
+
 lxc exec v1 -- /snap/bin/lxc launch "${TEST_IMG:-ubuntu-minimal-daily:24.04}" v1v1 --vm
 sleep 30
 lxc exec v1 -- /snap/bin/lxc info v1v1 | grep -F RUNNING
 
+if hasNeededAPIExtension devlxd_images && hasNeededAPIExtension event_lifecycle; then
```
simondeziel, line 130: Same comment re API extension support.

tests/devlxd-vm
```diff
@@ -90,14 +95,47 @@ fi
 # Wait for boot to be completed to interact with snap/snapd
 waitInstanceBooted v1
 
+# Below tests for image export rely on an event being emitted on export. This will not be backported to 4.0 or 5.0,
+# so exit the test here.
+if [ "${LXD_SNAP_CHANNEL}" = "4.0/edge" ] || [ "${LXD_SNAP_CHANNEL}" = "5.0/edge" ]; then
```
simondeziel, line 100: Could we bail out if `devlxd_images_vm` (when it's a thing) is not detected?

tests/devlxd-vm
markylaing: Yep, fixed :)

tests/devlxd-vm
markylaing: I've just removed this block so that the test continues. This way we're still testing that the nested VM works on older versions. I'm going to revisit the container tests to do the same thing.

simondeziel: Maybe, just remove the `event_lifecycle` to avoid running into a merge conflict.

tomponline:
ready for rebase

tests/devlxd-vm
markylaing: Done.

tomponline (approved):

tomponline:
@markylaing thanks!

devlxd-container seems to fail on edge intermittently https://github.com/canonical/lxd-ci/actions/runs/10280137408/job/28446865143?pr=255

PR #256, opened 2024-08-06
tests: Check that volume is functional after live migration.

markylaing:
Changes are as follows:
- Increases the memory of the nested VM to 1GiB (the parent VMs have 2GiB each).
- Waits indefinitely for the nested VM to start up by checking the number of process via `lxc info`.
- Formats the volume with ext4, mounts it, and adds a file.
- Checks that the file is present and has the same contents after live migrating the instance.


tomponline (dismissed):
Thanks!

markylaing:
Hmmm we'll it works locally but not in CI... what a surprise! :facepalm: 

I'll have to investigate more tomorrow with tmate.

tests/vm-migration
```diff
@@ -119,15 +136,33 @@ lxc exec member1 -- lxc storage create ceph ceph
...
 
 # Add vol1 as a disk device to the VM.
 lxc exec member1 -- lxc config device add v1 vol1-disk disk pool=ceph source=vol1
 
 # Start the VM.
 lxc exec member1 -- lxc start v1
-sleep 60
+
+# Wait for a long time for it to boot (doubly nested VM takes a while).
+while [ "$(lxc exec member1 -- sh -c "lxc info v1 | grep -F 'Processes:' | cut -d':' -f2 | tr -d '[:blank:]'")" -le 1 ]; do
```
simondeziel, line 148: I think the wrapper shell is not needed.

```suggestion
while [ "$(lxc exec member1 -- lxc info v1 | awk '{if ($1 == "Processes:") print $2}')" -le 1 ]; do
```
Also changed to use `awk` like we do elsewhere to get the process count.

tests/vm-migration
```diff
@@ -119,15 +136,33 @@ lxc exec member1 -- lxc storage create ceph ceph
...
+
+# vol1 should be available as /dev/sdb. Format it as ext4. Then mount it and create a file.
+lxc exec member1 -- sh -c "lxc exec v1 -- mkfs -t ext4 /dev/sdb"
+lxc exec member1 -- sh -c "lxc exec v1 -- mkdir /mnt/vol1"
+lxc exec member1 -- sh -c "lxc exec v1 -- mount -t auto /dev/sdb /mnt/vol1"
+lxc exec member1 -- sh -c "lxc exec v1 -- sh -c 'echo foo > /mnt/vol1/bar'"
 lxc exec member1 -- lxc move v1 --target member2
 
+# The VM is slow. So the agent isn't immediately available after the live migration.
+while [ "$(lxc exec member2 -- sh -c "lxc info v1 | grep -F 'Processes:' | cut -d':' -f2 | tr -d '[:blank:]'")" -le 1 ]; do
```
simondeziel, line 160: Same note about subshell.

tests/vm-migration
```diff
@@ -119,15 +136,33 @@ lxc exec member1 -- lxc storage create ceph ceph
...
+  sleep 5
+done
+
+# The volume should be functional, still mounted, and the file we created should still be there with the same contents.
+[ "$(lxc exec member2 -- sh -c "lxc exec v1 -- cat /mnt/vol1/bar")" = "foo" ]
```
simondeziel, line 165: This should work, I think:
```suggestion
[ "$(lxc exec member2 -- lxc exec v1 -- cat /mnt/vol1/bar)" = "v1" ]
```

tests/vm-migration
```diff
@@ -119,15 +136,33 @@ lxc exec member1 -- lxc storage create ceph ceph
...
+# Wait for a long time for it to boot (doubly nested VM takes a while).
+while [ "$(lxc exec member1 -- sh -c "lxc info v1 | grep -F 'Processes:' | cut -d':' -f2 | tr -d '[:blank:]'")" -le 1 ]; do
+  sleep 30
+done
+
+# vol1 should be available as /dev/sdb. Format it as ext4. Then mount it and create a file.
+lxc exec member1 -- sh -c "lxc exec v1 -- mkfs -t ext4 /dev/sdb"
+lxc exec member1 -- sh -c "lxc exec v1 -- mkdir /mnt/vol1"
+lxc exec member1 -- sh -c "lxc exec v1 -- mount -t auto /dev/sdb /mnt/vol1"
+lxc exec member1 -- sh -c "lxc exec v1 -- sh -c 'echo foo > /mnt/vol1/bar'"
```
simondeziel, line 156: Could we do without the subshells and also make it explicit that we expect an ext4 formatted disk rather than accept anything:

```suggestion
lxc exec member1 -- lxc exec v1 -- mkfs -t ext4 /dev/sdb
lxc exec member1 -- lxc exec v1 -- mkdir /mnt/vol1
lxc exec member1 -- lxc exec v1 -- mount -t ext4 /dev/sdb /mnt/vol1
lxc exec member1 -- lxc exec v1 -- cp /etc/hostname /mnt/vol1/bar
```

tests/vm-migration
markylaing: With awk I think I need the subshell:
```
$ [ lxc exec member1 -- lxc info v1 | awk '{if ($1 == "Processes:") print $2}' -le 1 ]
awk: fatal: cannot open file `-le' for reading: No such file or directory
```

tomponline:
ready for rebase

markylaing:
I'm pretty sure that the doubly nested VM isn't booting because the CPU usage is at 100% for qemu:
image

tests/vm-migration
simondeziel: Sorry for not being clear. With the subshell, I meant the `sh -c` bit used while doing the `lxc exec`.

tests/vm-migration
```diff
@@ -53,9 +53,16 @@ for _ in $(seq 60); do
...
-lxc launch "${TEST_IMG:-ubuntu-minimal-daily:24.04}" member1 --vm -c limits.memory=2GiB
-lxc launch "${TEST_IMG:-ubuntu-minimal-daily:24.04}" member2 --vm -c limits.memory=2GiB
+# Launch two instances for our LXD cluster and wait for them to be ready. If the host supports `devlxd_images_vm` then
+# set `security.devlxd.images=true` so that we don't have to download the image again.
+if hasNeededAPIExtension devlxd_images_vm; then
+  lxc launch "${TEST_IMG:-ubuntu-minimal-daily:24.04}" member1 --vm -c security.devlxd.images=true
+  lxc launch "${TEST_IMG:-ubuntu-minimal-daily:24.04}" member2 --vm -c security.devlxd.images=true
+else
+  lxc launch "${TEST_IMG:-ubuntu-minimal-daily:24.04}" member1 --vm
+  lxc launch "${TEST_IMG:-ubuntu-minimal-daily:24.04}" member2 --vm
```
simondeziel, line 63: The GHA runner VM have 16G of RAM (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) so you can probably allocate more RAM to those memberX VMs.

Have you already tried with `-c limits.cpu=2 -c limits.memory=4GiB`?

markylaing: I've bumped all the resources to what is used in microcloud and it's still not booting after over an hour. I'll go back to trying with containers and see how I get on.

tomponline: yes this would avoid needing to keep setting up the websockets over vsock for the exec sessions

simondeziel:
@markylaing it seems the `vm-migration` test is using a `dir` pool. The microcloud test suite uses a `zfs` one which might make it faster to boot the many VMs without having to hit the disk as hard due to CoW taking place. Maybe it's worth trying `zfs` here as well?

tomponline:
@markylaing @simondeziel any progress on using container cluster members for ceph backed VMs?

simondeziel:
@tomponline it's in my queue for this pulse, but no progress as of yet.

MusicDin:
FYI: I did some tests with @mihalicyn, and even using a local storage without any attached custom volumes (no Ceph involved) is not sufficient to start a VM on third layer of nested virtualization (GH runner > LXD VM > LXD VM).

MusicDin:
I'm wondering whether there is a way to spawn two LXDs on a GH runner and connect them into a cluster to avoid using 1 layer of virtualization.

Otherwise, I assume the testflinger is the way to go

PR #257, opened 2024-08-06
tests/devlxd-container: devlxd_images and event_lifecycle are supported by LXD 4.0+

simondeziel:
Stop testing for extensions supported by the oldest LXD version still supported.


simondeziel:
Relevant tests all green.

tomponline (approved):

PR #263, opened 2024-08-13
tests/network: switch from 23.10 (EOL) to 24.04 (LTS)

simondeziel:


simondeziel:
@tomponline `tests/network` is not tested on GHA (so all red in CI is to be ignored) as it requires SR-IOV NICs. We usually run it through testflinger (on `luma`) but testflinger is in downtime this week.

tomponline (approved):
We should get the non physical tests working in github too.

simondeziel:
> We should get the non physical tests working in github too.

Just to be 100% sure, you are referring to the macvlan ones, right?

tomponline:
Yeah anything that doesn't require a physical nic

PR #265, opened 2024-08-14
Sleep after hotplugging block dev

hamistao:
This test has shown to be flaky because the device is not effectively added before the next command executes (See https://github.com/canonical/lxd-ci/actions/runs/10297255515/job/28500067002#step:8:637).
So a short sleep makes this test more consistent, as it is already done in other places in the tests where hotplugging devices is performed.


tomponline (approved):
Ta

PR #266, opened 2024-08-16
tests/storage-disks-vm: Add tests for virtio-blk and virtio-scsi

kadinsayani:
Adds tests for canonical/lxd#13905

### DIFF
--- a/images/springdalelinux.yaml
+++ b/images/springdalelinux.yaml
@@ -193,7 +193,7 @@ files:
 - path: /etc/dracut.conf.d/lxd.conf
   generator: dump
   content: |-
-    add_drivers+=" virtio_scsi virtio_pci virtio_console "
+    add_drivers+=" virtio_scsi virtio_pci virtio_console virtio_blk "
   types:
   - vm
   releases:
@@ -202,7 +202,7 @@ files:
 - path: /etc/dracut.conf.d/lxd.conf
   generator: dump
   content: |-
-    add_drivers+=" virtio_scsi virtio_console sd_mod "
+    add_drivers+=" virtio_scsi virtio_console sd_mod virtio_blk "
   types:
   - vm
   releases:
--- a/tests/storage-disks-vm
+++ b/tests/storage-disks-vm
@@ -33,9 +33,19 @@ truncate -s 5m "${testRoot}/allowed1/lxd-block-test"
 touch "${testRoot}/allowed1/lxd-test"
 
 # Create loop device to use as NVMe disk
-loopimg="$(mktemp)"
-truncate -s 100M "${loopimg}"
-loopdev="$(losetup --show --find "${loopimg}")"
+loopImgNVME="$(mktemp --suffix -nvme)"
+truncate -s 100M "${loopImgNVME}"
+loopDevNVME="$(losetup --show --find "${loopImgNVME}")"
+
+# Create loop device to use as virtio-blk disk
+loopImgVirtioBlk="$(mktemp --suffix -virtio-blk)"
+truncate -s 100M "${loopImgVirtioBlk}"
+loopDevVirtioBlk="$(losetup --show --find "${loopImgVirtioBlk}")"
+
+# Create loop device to use as virtio-scsi disk
+loopImgVirtioSCSI="$(mktemp --suffix -virtio-scsi)"
+truncate -s 100M "${loopImgVirtioSCSI}"
+loopDevVirtioSCSI="$(losetup --show --find "${loopImgVirtioSCSI}")"
 
 # Configure LXD
 lxc project switch default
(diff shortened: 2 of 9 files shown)


tomponline:
Shall we add some for the nvme mode too whilst we are at it?

kadinsayani:
> Shall we add some for the nvme mode too whilst we are at it?

The code initially included a test for nvme, so we're all good there 😄

images/springdalelinux.yaml
```diff
@@ -202,7 +202,7 @@ files:
 - path: /etc/dracut.conf.d/lxd.conf
   generator: dump
   content: |-
-    add_drivers+=" virtio_scsi virtio_console sd_mod "
```
simondeziel, line 205: I'm not 100% sure of what dracut does but it looks like it's asking for the virtio-blk driver to be bundled in the initramfs even if not detect at initramfs build time. If that's the case, nice catch!

That said, I'm not entirely sure if we should do that and at least we didn't do it for NVME disks either and I'm not aware of anyone complaining. Using a non-default driver is probably an edge case so maybe it's OK to expect users making that choice to do some customization to the image they want to use?

To be honest though, bundling that driver might be really quick and cheap? I couldn't check the size of the said module on Ubuntu as it's builtin.

@tomponline opinions?

tests/storage-disks-vm
```diff
@@ -33,9 +33,19 @@ truncate -s 5m "${testRoot}/allowed1/lxd-block-test"
 touch "${testRoot}/allowed1/lxd-test"
 
 # Create loop device to use as NVMe disk
-loopimg="$(mktemp)"
-truncate -s 100M "${loopimg}"
-loopdev="$(losetup --show --find "${loopimg}")"
+loopimg1="$(mktemp)"
+truncate -s 100M "${loopimg1}"
+loopdev1="$(losetup --show --find "${loopimg1}")"
```
simondeziel, line 38: For ease of debugging (when `set -x`):
```suggestion
loopNVME="$(mktemp --suffix -nvme)"
truncate -s 100M "${loopNVME}"
loopNVME="$(losetup --show --find "${loopNVME}")"
```

tests/storage-disks-vm
```diff
@@ -322,24 +332,37 @@ if lxc exec v1 -- mount | grep -qwF /boot/efi; then
...
   lxc project unset restricted restricted.devices.disk.paths
   lxc project unset restricted restricted
 
-  # Add a NVMe disk and check if a NVMe controller is added to the VM
-  lxc config device add v1 nvme-ssd disk source="${loopdev}" io.bus=nvme
+  # Add a NVMe disk
+  lxc config device add v1 nvme-ssd disk source="${loopdev1}" io.bus=nvme
+  
+  # Add a virtio-blk disk
+  lxc config device add v1 virtio-blk-ssd disk source="${loopdev2}" io.bus=virtio-blk
```
simondeziel, line 346: This needs to be gated by `hasNeededAPIExtension disk_io_bus_virtio_blk` as we want the same script to work on all supported LXD versions.

tests/storage-disks-vm
```diff
@@ -322,24 +332,37 @@ if lxc exec v1 -- mount | grep -qwF /boot/efi; then
...
+  waitInstanceReady v1 
+
+  # Check if NVMe, virtio-blk and virtio-scsi controllers are added to the VM
   if ! lxc exec v1 -- sh -c 'command -v lspci'; then
     lxc exec v1 -- apt-get update
     lxc exec v1 -- apt-get install --no-install-recommends -y pciutils
   fi
   lxc exec v1 -- lspci | grep -F "QEMU NVM Express Controller"
+  lxc exec v1 -- lspci | grep -F "Red Hat, Inc. Virtio block device"
+  lxc exec v1 -- lspci | grep -F "Red Hat, Inc. Virtio SCSI"
```
simondeziel, line 362: In my quick test on a 24.04 VM, I got a `1.0` getting in the way, preventing that `grep` from matching:

```
root@v1:~# lspci | grep -F "Red Hat, Inc. Virtio" | grep -wF SCSI
02:00.0 SCSI storage controller: Red Hat, Inc. Virtio 1.0 SCSI (rev 01)
```

It's probably the same for the `block device` one which would explain why the tests are failing.

tests/storage-disks-vm
```diff
@@ -322,24 +332,37 @@ if lxc exec v1 -- mount | grep -qwF /boot/efi; then
...
+  # Start VM
   lxc start v1
-  waitInstanceReady v1
+  waitInstanceReady v1 
+
+  # Check if NVMe, virtio-blk and virtio-scsi controllers are added to the VM
   if ! lxc exec v1 -- sh -c 'command -v lspci'; then
     lxc exec v1 -- apt-get update
     lxc exec v1 -- apt-get install --no-install-recommends -y pciutils
   fi
```
simondeziel, line 359: For debuggability, we should add a `lxc exec v1 -- lspci` right here. This way, we'd better understand why the `grep`s are not matching.

simondeziel (commented):
Nice work! Just a few nits to address please.

images/springdalelinux.yaml
kadinsayani: > I'm not 100% sure of what dracut does but it looks like it's asking for the virtio-blk driver to be bundled in the initramfs even if not detect at initramfs build time. If that's the case, nice catch!

dracut creates an initramfs image and bundles the specified drivers which aren't included on these distros by default.

> That said, I'm not entirely sure if we should do that and at least we didn't do it for NVME disks either and I'm not aware of anyone complaining. Using a non-default driver is probably an edge case so maybe it's OK to expect users making that choice to do some customization to the image they want to use?

I believe NVMe is included by default for these distros, so we don't need to include it in ```dracut.conf```. I'm not 100% sure on this though.

> To be honest though, bundling that driver might be really quick and cheap? I couldn't check the size of the said module on Ubuntu as it's builtin.

I was able to check the size on one of my VMs, see below for reference:

```lsmod | grep virtio_blk | awk '{print $1, $2/1024 " MB"}'
virtio_blk 20 MB```

After reading through the dracut docs and trying to understand how it works, it appears to be quite efficient at generating a customized initramfs image. With that being said, it doesn't hurt to investigate/test this further.

tomponline (approved):
LGTM thanks!

Nice thought about adding support for this mode to our images too. Well done!

PR #267, opened 2024-08-19
OpenWrt 22.03 is EOL

simondeziel:
Release 22.03 went EOL in July 2024, https://openwrt.org/docs/guide-developer/security#support_status


MusicDin (approved):
Thanks. Nice catch with name capitalization :)

tomponline (approved):

PR #270, opened 2024-08-19
github: Void Linux only has a current release

simondeziel:
There doesn't seem to be any 23.11 release: https://mirrors.servercentral.com/voidlinux/live/

It has 2 variants (`default` (glibc) and `musl`) but we only provide the `default` one.

### DIFF
--- a/.github/workflows/image-voidlinux.yml
+++ b/.github/workflows/image-voidlinux.yml
@@ -23,8 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         release:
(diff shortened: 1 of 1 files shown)


MusicDin:
Should we also include `musl`?

MusicDin (approved):

simondeziel:
> Should we also include `musl`?

I thought of this but there doesn't seem to much demand for Void Linux images.

tomponline (approved):

PR #272, opened 2024-08-20
tests/cgroup: on cgroup1 systems, the swap limit is set to limits.memory * 2

simondeziel:


simondeziel:
As mentioned in chat, I think that trying to make all tests work on 20.04 with modern LXD is going to be a time sync. Expanding the exclusion matrix to limit what is tested on 20.04 is probably the way to go (rather than merging this PR).

tomponline:
Lets wait until @mihalicyn has a look, if this is expected behaviour then we can indeed test for it.

tomponline:
@mihalicyn @simondeziel do we still need this?

tomponline:
Any news chaps @mihalicyn @simondeziel ?

simondeziel:
This has stopped being a pain point in CI because we no longer test 5.0 on `cgroup1` systems (`20.04`). I'm not sure if the PR does the right thing (accept different limits based on the LXD version) or if LXD should always use the same limit between every versions.

I guess it's a bit late to change LXD's behaviour now so this PR should be considered if we want to start testing `5.0` on `20.04` again.

PR #273, opened 2024-08-20
github: allow manually running tests on specific Ubuntu releases

simondeziel:


simondeziel:
@tomponline this should be handy when you want to schedule tests against say `5.21/candidate` which is known to not play well with 20.04 for example. See it live at https://github.com/simondeziel/lxd-ci/actions/runs/10478318542

tomponline (approved):
Ta

PR #274, opened 2024-08-21
tests: Add virtio conversion test

MusicDin:


tests/conversion
```diff
@@ -5,6 +5,7 @@ set -eu
 . ./bin/helpers
 
 install_deps jq unzip
+install_deps virt-v2v || true
```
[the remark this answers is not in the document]
MusicDin: Should have added the comment. This package is not available on an older 20.04 runner. virt-v2v was part of libguestfs-utils previously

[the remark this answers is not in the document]
tomponline: ok, please can you tweak this so that we check if package is available before installing it, that way if package install fails for some other reason we wont skip the test further down and will fail instead.

tests/conversion
```diff
@@ -272,6 +282,41 @@ if hasNeededAPIExtension instance_import_conversion; then
...
+        # Ensure instance does not boot without virtio_scsi drivers.
+        lxc start v1
+        ! waitInstanceReady v1 || { echo "Instance booted without SCSI drivers!"; false; }
+        lxc delete -f v1
+
+        # Extract raw image.
+        tar -xzf "${tmpdir}/vm.tar.gz" -C "${tmpdir}" "backup/virtual-machine.img"
+        rm "${tmpdir}/vm.tar.gz"
+
+        # Ensure that virtio conversion correctly rebuilds the initramfs.
```
simondeziel, line 316:
```suggestion
        # Ensure that virtio conversion adds the needed virtio_scsi module back into the initramfs.
```

tests/conversion
```diff
@@ -6,6 +6,13 @@ set -eu
 
 install_deps jq unzip
 
+# Install virt-v2v. On older Ubuntu versions, this package is available
+# under libguestfs-tools, but we do not want that as virt-v2v version
+# is too old anyway.
+if apt-cache show virt-v2v > /dev/null; then
```
simondeziel, line 12: `apt-cache show` only works if a prior `apt-get update` was run. Since you've called `install_deps` before, you have it implicitly but that feels brittle.

As such, I'd prefer if we hardcode the installation when:

```
# XXX: virt-v2v is recent enough in 24.04 onward. For older releases, simply skip the v2v conversion test.
. /etc/os-release
if dpkg --compare-versions "${VERSION_ID}" ge 24.04; then
      install_deps jq unzip virt-v2v
else
      install_deps jq unzip
fi
```

tests/conversion
```diff
@@ -272,6 +282,41 @@ if hasNeededAPIExtension instance_import_conversion; then
     conversion_vm vm-u24-vmdk-dir-bupvol dir "${IMAGE_PATH}" "yes"
     conversion_vm vm-u24-vmdk-zfs-bupvol zfs "${IMAGE_PATH}" "yes"
 
+    # Test virtio conversion if virt-v2v-in-place is installed and meets the minimum version
+    # requirements. Version will be satisfied from Ubuntu 24.04 runners onwards.
+    v2vMinVersion="2.3.4"
```
simondeziel, line 287: While this command/version detection logic seems sound, I worry that it's a bit brittle and we risk skipping the v2v conversion if it doesn't work.

tests/conversion
```diff
@@ -272,6 +282,41 @@ if hasNeededAPIExtension instance_import_conversion; then
...
+    if ! command -v virt-v2v-in-place > /dev/null; then
```
simondeziel, line 288:
```suggestion
    if ! dpkg --compare-versions "${VERSION_ID}" ge 24.04 then
```

simondeziel (commented):
LGTM with some small comments for improved robustness

tests/conversion
MusicDin: With your suggestion above to check the OS version instead of virt-v2v version, this is no longer required. If the installed version is for any reason lower then 2.3.4, LXD will throw an intuitive error, so it should be easy to catch.

tests/conversion
MusicDin: Thanks, a very nice way of loading release variables that I never thought about :)

tests/conversion
MusicDin: I suppose the `command` is fine now, considering we no longer check for the virt-v2v version

simondeziel (approved):

tomponline (approved):
Excellent thanks

PR #277, opened 2024-08-23
Don't use `--accept-certificate` when consuming token

simondeziel:
The whole point of using token is to avoid blindly trusting a random server certificate. IMHO, LXD should refuse this combo (https://github.com/canonical/lxd/issues/13973).


simondeziel:
@tomponline the failures are unrelated (known flakeyness in the tests) so "all green".

tomponline (approved):

PR #278, opened 2024-08-27
Use `bash` for test scripts

simondeziel:


simondeziel:
@tomponline no regression despite not being 100% green.

tomponline (approved):

PR #280, opened 2024-08-28
tests/container: add checks for procfs/devpts mounts

mihalicyn:
Check that we can mount devpts and procfs inside an unprivileged container.

This tests for the following changes:
https://github.com/canonical/lxd/commit/6d6aad0d1c79344c019a0997077a0c502bed1da3
https://github.com/canonical/lxd/commit/861923942f162b8ab7e92125a97455ea1d4d2468


tests/container
```diff
@@ -135,5 +135,15 @@ for release in $RELEASES; do
...
 
+IMAGE="ubuntu-minimal-daily:24.04"
+
+echo "==> check that we can mount devpts and procfs in unprivileged container (${release})"
+lxc launch "${IMAGE}" u1
+lxc exec u1 -- mkdir /root/proc
+lxc exec u1 -- mount -t proc proc /root/proc
+lxc exec u1 -- mkdir /root/devpts
+lxc exec u1 -- mount -t devpts devpts /root/devpts
+lxc delete -f u1
```
simondeziel, line 146: Could we have that tested for 22.04 and 24.04 maybe? If so, could that be moved into the for loop?

mihalicyn: It's up to you. I thought that as these checks and logic behind it completely distro-independent it's better to have it out of the loop to make test a bit faster.

simondeziel: Works for me but then `$release` is used out of scope (I know that's not really it but you get the point).

mihalicyn: ah, yeah. Fixed! Thanks!

tomponline:
thanks @mihalicyn !

PR #281, opened 2024-08-28
Storage bit-for-bit comparison

simondeziel:


simondeziel:
@tomponline all green for the `tests/storage-vm` test against all snap channels. The execution time is ~30s longer than before.

tomponline (approved):
Thanks!

PR #283, opened 2024-09-04
Add tests for non-interactive lxd-migrate commands

MusicDin:
Add tests for non-interactive usage of `lxd-migrate` (using flags).

### DIFF
--- a/tests/conversion
+++ b/tests/conversion
@@ -87,6 +87,31 @@ checkDNS() {
     return 1
 }
 
+testImportedInstance() {
+    instName=$1
+
+    # Start the instance.
+    echo "Starting instance ${instName}..."
+    lxc start "${instName}"
+
+    # Wait for the instance to be ready and ensure it has global ip address.
+    echo "Waiting instance ${instName} to start..."
+    waitInstanceReady "${instName}"
+    sleep 10
+
+    # Print the instances.
+    lxc list
+
+    echo "Check network connectivity of instance ${instName}..."
+    checkIPAddresses "${instName}"
+
+    echo "Check DNS resolution of instance ${instName}..."
+    checkDNS "${instName}"
+
+    # Cleanup.
+    lxc delete -f "${instName}"
+}
+
 # conversion_vm creates a new LXD virtual machine from the image on the given path.
 # A dedicated storage pool of the given type will be created for a new virtual machine.
 #
@@ -201,26 +226,7 @@ conversion() {
         echo "1"
     } | $lxdMigrateCmd
 
-    # Start the instance.
-    echo "Starting instance ${instName}..."
-    lxc start "${instName}"
-
-    # Wait for the instance to be ready and ensure it has global ip address.
-    echo "Waiting instance ${instName} to start..."
-    waitInstanceReady "${instName}"
-    sleep 10
-
-    # Print the instances.
-    lxc list
-
-    echo "Check network connectivity of instance ${instName}..."
-    checkIPAddresses "${instName}"
-
-    echo "Check DNS resolution of instance ${instName}..."
-    checkDNS "${instName}"
-
-    # Cleanup.
-    lxc delete -f "${instName}"
+    testImportedInstance "${instName}"
 
     if [ "${poolName}" != "" ]; then
         lxc storage delete "${poolName}"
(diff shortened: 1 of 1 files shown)


tests/conversion
```diff
@@ -247,13 +251,86 @@ lxc delete vm-tmp
...
+token="$(lxc config trust add --quiet --name migrate)"
+if [ -z "${token}" ]; then
+    echo "Failed to generate LXD trust token!"
+    return 1
+fi
+
+# Target server test.
+lxd-migrate --non-interactive --instance-type vm --source "${IMAGE_PATH}" --instance-name vm-ni-unix-default \
+    --config security.secureboot=false
+testImportedInstance "${instName}"
```
simondeziel, line 275: `s/${instName}/vm-ni-unix-default/` ?

tests/conversion
```diff
@@ -247,13 +251,86 @@ lxc delete vm-tmp
...
 # conversion API extension, lxd-migrate must fallback to migration
 # mode and successfully transfer the VM disk.
-conversion_vm vm-alpine-raw-btrfs btrfs "${tmpdir}/backup/virtual-machine.img" "no"
-conversion_vm vm-alpine-raw-lvm lvm "${tmpdir}/backup/virtual-machine.img" "no"
-conversion_vm vm-alpine-raw-zfs zfs "${tmpdir}/backup/virtual-machine.img" "no"
-conversion_vm vm-alpine-raw-dir dir "${tmpdir}/backup/virtual-machine.img" "no"
+conversion_vm vm-alpine-raw-btrfs btrfs "${IMAGE_PATH}" "no"
+conversion_vm vm-alpine-raw-lvm lvm "${IMAGE_PATH}" "no"
+conversion_vm vm-alpine-raw-zfs zfs "${IMAGE_PATH}" "no"
+conversion_vm vm-alpine-raw-dir dir "${IMAGE_PATH}" "no"
```
simondeziel, line 262: How about a `for driver in btrfs dir lvm zfs; do` loop?

tests/conversion
MusicDin: Agree, code would look nicer. I prefer having different names just because it is easier to find the relevant part in script if something goes wrong :)

MusicDin: I see, thanks!

tests/conversion
```diff
@@ -247,13 +253,99 @@ lxc delete vm-tmp
...
+    --instance-name vm-ni-custom-config \
+    --no-profiles \
+    --network lxdbr0 \
+    --storage default \
+    --storage-size 5GiB \
+    --config limits.cpu=2 \
+    --config limits.memory=2GiB \
+    --config security.secureboot=false
+query="$(lxc query /1.0/instances/vm-ni-custom-config)"
+[ "$(echo "${query}" | jq .profiles)" = "[]" ]                   # Ensure no profiles are applied.
```
simondeziel, line 310: `jq` without `-r` sticks out here :)

tests/conversion
```diff
@@ -247,13 +253,99 @@ lxc delete vm-tmp
...
+    echo "Failed to generate LXD trust token!"
+    return 1
+fi
+
+# Test VM import with no target server - local unix server is used by default.
+lxd-migrate --non-interactive --instance-type vm --source "${IMAGE_PATH}" \
+    --instance-name vm-ni-unix-default \
+    --config security.secureboot=false
+query="$(lxc query /1.0/instances/vm-ni-unix-default)"
+[ "$(echo "${query}" | jq -r '.profiles[]')" = "default" ]       # Ensure default profile is applied.
```
simondeziel, line 278: As we use `bash` now, you can simplify this to:

```suggestion
[ "$(jq -r '.profiles[]' <<< "${query}")" = "default" ]           # Ensure default profile is applied.
```

It's optionally, really. The benefit is that it makes `set -x` runs less verbose while still being informative.

simondeziel (approved):
LGTM with small optional nitpicks

simondeziel:
@MusicDin I'm seeing the alpine VM test transferring a 10GiB volume:

```
Please pick one of the options above [default=1]: Please provide the storage pool to use: Do you want to change the storage volume size? [default=no]: Please specify the storage volume size: 
Instance to be created:
  Name: vm-alpine-raw-zfs
  Project: default
  Type: virtual-machine
  Source: /tmp/tmp.8z8X457HPK/backup/virtual-machine.img
  Storage pool: conversion-zfs
  Storage volume size: 10GiB
```

For Alpine, could we tune that down to say 4GiB? That's the smallest working size I could get from `images:`:

```
$ lxc launch images:alpine/edge -c security.secureboot=false --vm a1 -d root,size=1GiB
Launching a1
Error: Failed instance creation: Failed creating instance from image: Source image size (4294967296) exceeds specified volume size (1073741824)
$ lxc launch images:alpine/edge -c security.secureboot=false --vm a1 -d root,size=4GiB
Launching a1
```

Speaking of that 4GiB min size, I think we could even reduce that to 1GiB:

```
$ lxc exec a1 -- df -h
Filesystem                Size      Used Available Use% Mounted on
devtmpfs                 10.0M         0     10.0M   0% /dev
shm                     483.1M         0    483.1M   0% /dev/shm
/dev/sda2                 3.7G    248.9M      3.4G   7% /
tmpfs                   193.3M    264.0K    193.0M   0% /run
efivarfs                255.9K     20.6K    230.3K   8% /sys/firmware/efi/efivars
/dev/sda1                98.4M    282.0K     98.1M   0% /boot/efi
tmpfs                    50.0M     12.5M     37.5M  25% /run/lxd_agent
```

tomponline (approved):

PR #284, opened 2024-09-06
Tests for disconnections during `lxc exec`

hamistao:
As was discussed with @mihalicyn If containers are stopped gracefully, the exit code can be 129(SIGHUP) or 143(SIGTERM) depending on the signal used, both results have been seen and make sense. If forcefully stopping a container the exit code is always 137 (SIGKILL). For VMs, the exit code was standardized as 129 for all cases by [#13875](https://github.com/canonical/lxd/pull/13875) (not merged yet).

In its current state, this PR only adds tests for forcefully stopping containers and VMs.

### DIFF
--- a/tests/container
+++ b/tests/container
@@ -139,6 +139,7 @@ IMAGE="ubuntu-minimal-daily:24.04"
 
 echo "==> check that we can mount devpts and procfs in unprivileged container"
 lxc launch "${IMAGE}" u1
+waitInstanceBooted u1
 lxc exec u1 -- mkdir /root/proc
 lxc exec u1 -- mount -t proc proc /root/proc
 lxc exec u1 -- mkdir /root/devpts
--- a/tests/container
+++ b/tests/container
@@ -147,5 +147,13 @@ lxc exec u1 -- mkdir /root/sys
 lxc exec u1 -- mount -t sysfs sysfs /root/sys
 lxc delete -f u1
 
+
+echo "==> Try cleanly stopping a container"
+lxc launch "${IMAGE}" c1
+waitInstanceBooted c1
+lxc stop c1
+lxc start c1
+waitInstanceBooted c1
+
 # shellcheck disable=SC2034
 FAIL=0
--- a/tests/container
+++ b/tests/container
@@ -147,13 +147,30 @@ lxc exec u1 -- mkdir /root/sys
 lxc exec u1 -- mount -t sysfs sysfs /root/sys
 lxc delete -f u1
 
-
 echo "==> Try cleanly stopping a container"
 lxc launch "${IMAGE}" c1
 waitInstanceBooted c1
 lxc stop c1
 lxc start c1
 waitInstanceBooted c1
+
+echo "==> Test exit codes when container disconnects during lxc exec"
+
+# Try disconnecting a container stopping forcefully and gracefully to make sure they differ appropriately.
+(sleep 1 && lxc stop -f c1) &
+lxc exec c1 -- sleep 10 || exitCode=$?
+[ "${exitCode:-0}" -eq 137 ]
+
+wait $!
+lxc start c1
+waitInstanceBooted c1
+(sleep 1 && lxc stop c1) &
+lxc exec c1 -- sleep 10 || exitCode=$?
+# Both 129 and 143 have been seen and both make sense here.
+[ "${exitCode:-0}" -eq 129 ] || [ "${exitCode:-0}" -eq 143 ]
+
+wait $!
+lxc delete -f c1
 
 # shellcheck disable=SC2034
 FAIL=0
--- a/tests/vm
+++ b/tests/vm
@@ -16,6 +16,25 @@ waitInstanceBooted vm1
 echo "==> Test cleanly stopping a VM"
 lxc stop vm1
 
+if ! echo "${LXD_SNAP_CHANNEL}" | grep -qE '^4\.0/'; then
+    echo "==> Test lxc exec exit code upon VM disconnection due to a stop/reboot"
+    lxc start vm1
+    waitInstanceBooted vm1
+
+    # Try disconnecting a VM stopping forcefully and gracefully to make sure they match.
+    (sleep 1 && lxc stop -f vm1) &
+    lxc exec vm1 -- sleep 10 || exitCode=$?
+    [ "${exitCode:-0}" -eq 129 ]
+
+    wait $!
+    lxc start vm1
+    waitInstanceBooted vm1
+    (sleep 1 && lxc stop vm1) &
+    lxc exec vm1 -- sleep 10 || exitCode=$?
+    [ "${exitCode:-0}" -eq 129 ]
+    wait $!
+fi
+
 # Cleanup
 lxc delete -f vm1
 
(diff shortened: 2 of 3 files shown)


simondeziel:
There are too many commits in here, @hamistao did something went wrong when pushing/opening the PR?

hamistao:
@simondeziel Yes I messed up my rebase, I should have opened this as a draft. Sorry about that.

hamistao:
@simondeziel Fixed, this should not be merged yet since the VM fix tested here is not in LXD yet, but feel free to take a look whenever you can.

simondeziel:
> @simondeziel Yes I messed up my rebase, I should have opened this as a draft. Sorry about that.

No worries at all!

tests/container
```diff
@@ -147,5 +147,14 @@ lxc exec u1 -- mkdir /root/sys
 lxc exec u1 -- mount -t sysfs sysfs /root/sys
 lxc delete -f u1
 
+echo "==> Test exit codes when container disconnects during lxc exec"
+lxc launch "${IMAGE}" c1
+
+(sleep 1 && lxc stop -f c1) &
+lxc exec c1 -- sleep 10 || exitCode=$?
+[ "${exitCode:-0}" -eq 129 ]
```
simondeziel, line 155: Hey, I was about to comment that checking for `SIGKILL` was potentially risky as I'd expect `SIGTERM` to also be a possible return code, no?

My assumption is that `lxc stop` sends a `SIGTERM` first, wait a bit then send a `SIGTERM`. Assuming that assumption is valid, would it be possible the first `SIGTERM` is enough to stop everything cleanly in the instance and thus change the return code?

Also, I'd probably want that `128+SIG_NUMBER_` logic briefly mentioned here cause `SIGKILL=9` and `SIGTERM=15` is widely known but `129` or `137` are more like magic numbers to me ;)

tests/vm
```diff
@@ -0,0 +1,25 @@
...
+install_lxd
+
+# Configure LXD
+lxd init --auto
+
+IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
+vmName="test-vm"
+
+# Launch test instance
+lxc launch "${IMAGE}" "${vmName}" --vm
```
simondeziel, line 14: I'd go with `v1` hardcoded everywhere to make it easier to read through (when it grows longer). This also better fit with other test scripts IMHO.

tests/container
hamistao: Sure! But could you please elaborate on the "And what about a graceful stop not aborting anything" case?

simondeziel: > But could you please elaborate on the "And what about a graceful stop not aborting anything" case?

Just a plain, `lxc launch`, wait ready then `lxc stop`.

tests/container
```diff
@@ -147,5 +147,10 @@ lxc exec u1 -- mkdir /root/sys
 lxc exec u1 -- mount -t sysfs sysfs /root/sys
 lxc delete -f u1
 
+
+echo "==> Try cleanly stopping a container"
```
tomponline, line 151: This is going to be unreliable as the guest OS wont have booted yet and wont respond

tests/container
```diff
@@ -147,5 +148,25 @@ lxc exec u1 -- mkdir /root/sys
...
 lxc delete -f u1
 
+echo "==> Try cleanly stopping a container"
+lxc launch "${IMAGE}" c1
+waitInstanceBooted c1
+lxc restart c1
+
+echo "==> Test exit codes when container disconnects during lxc exec"
+
+# Try disconnecting a container stopping forcefully and gracefully to make sure they differ appropriately.
```
simondeziel, line 158: After starting `c1`, I'd probably `waitInstanceBooted c1` before doing a `(sleep 1 && lxc stop -f c1)`.

tests/container
```diff
@@ -147,5 +148,25 @@ lxc exec u1 -- mkdir /root/sys
...
+(sleep 1 && lxc restart -f c1) &
```
simondeziel, line 159: I'd use `lxc stop -f` here too.

tests/vm
```diff
@@ -0,0 +1,34 @@
...
+lxd init --auto
+
+IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
+
+# Launch test instance
+lxc launch "${IMAGE}" vm1 --vm
+
+echo "==> Try cleanly stopping a VM"
+waitInstanceReady vm1
+lxc restart vm1
```
simondeziel, line 17: Same comment re `restart` vs `stop`+`start`.

hamistao:
@simondeziel This is ready for another review

tests/vm
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
simondeziel, line 1: As a new test, it needs to be added to `.github/workflows/tests.yml` in the long list of tests we run. Please put it alphabetically sorted in the list.

simondeziel (changes requested):

hamistao:
@simondeziel I keep getting a `context deadline exceeded error` in the VM clean stop tests.
If you agree, I suggest removing those tests and open this for another review.

tests/vm
```diff
@@ -0,0 +1,41 @@
...
+install_lxd
+
+# Configure LXD
+lxd init --auto
+
+IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
+
+# Launch test instance
+lxc launch "${IMAGE}" vm1 --vm
+waitInstanceReady vm1
```
simondeziel, line 14:
```suggestion
waitInstanceBooted vm1
```

tests/vm
```diff
@@ -0,0 +1,41 @@
...
+
+echo "==> Test cleanly stopping a VM"
+lxc stop vm1
+lxc start vm1
+waitInstanceReady vm1
```
simondeziel, line 19:
```suggestion
waitInstanceBooted vm1
```

simondeziel (commented):
https://github.com/canonical/lxd-ci/actions/runs/10885376800/job/30202865680?pr=284 taking ~2h feels like a bug as I'd expect `lxc stop` to timeout before that long, no?

That said, I think the context deadline bit might be addressed by waiting for complete boot before doing the shutdown.

tests/vm
hamistao: > taking ~2h feels like a bug as I'd expect lxc stop to timeout before that long, no?

I think the timeout is indeed very long so this should be expected

> That said, I think the context deadline bit might be addressed by waiting for complete boot before doing the shutdown.

Yeah, that makes sense, thanks!

hamistao:
@simondeziel This is ready again

tests/vm
```diff
@@ -0,0 +1,41 @@
...
+IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
+
+# Launch test instance
+lxc launch "${IMAGE}" vm1 --vm
+waitInstanceBooted vm1
+
+echo "==> Test cleanly stopping a VM"
+lxc stop vm1
+lxc start vm1
+waitInstanceBooted vm1
```
simondeziel, line 19: I think this could be moved inside the `if` as there's no point in starting the VM back if we are to delete it right after it's done booting.

simondeziel (approved):

tomponline (approved):

PR #287, opened 2024-09-12
tests/qemu-external-vm: rework version based skipping

simondeziel:
Testing only on `latest/edge` means that we'll need to revisit the condition when the next LTS occurs. Let's do it now as we know what to skip already.


simondeziel:
@tomponline the failures are unrelated.

tomponline (approved):

PR #293, opened 2024-09-23
tests/conversion: conserve CPU/disk by exporting to stdout

simondeziel:
After chatting about #292 with @MusicDin he mentioned that I forgot about this other case.


MusicDin (approved):
Thanks

simondeziel:
The first (alpine) one shaved ~35s but this one is even better as it goes from 147s down to 40s.

PR #294, opened 2024-09-23
tests/container: Add test to ensure launching an aliased container image won't launch with vm flag

kadinsayani:
- Launching an aliased container image shouldn't launch with the `--vm` flag; and
- VM's should be launched with `--vm` flag.

### DIFF
--- a/tests/container
+++ b/tests/container
@@ -155,6 +155,10 @@ lxc stop c1
 lxc start c1
 waitInstanceBooted c1
 
+echo "==> Ensure aliased image won't launch with VM flag set"
+lxc image alias create containeralias "$(lxc config get c1 volatile.base_image)"
+! lxc launch containeralias --vm || false
+
 echo "==> Test exit codes when container disconnects during lxc exec"
 
 # Try disconnecting a container stopping forcefully and gracefully to make sure they differ appropriately.
--- a/tests/storage-vm
+++ b/tests/storage-vm
@@ -612,7 +612,7 @@ for poolDriver in $poolDriverList; do
         lxc config set storage.images_volume "${poolName}"/images
         lxc publish v1 --alias v1image
         lxc delete v1
-        lxc launch v1image v2 -s "${poolName}"
+        lxc launch v1image v2 -s "${poolName}" # --vm flag omitted as we are implicitly launching a vm
         waitInstanceReady v2
         lxc delete -f v2
         lxc image delete v1image


tests/vm
```diff
@@ -16,6 +16,11 @@ waitInstanceBooted vm1
 echo "==> Test cleanly stopping a VM"
 lxc stop vm1
 
+# Ensure that we can't launch an aliased VM image as a container
```
simondeziel, line 19: Could we more simply get the `volatile.base_image` value from `v1`?

simondeziel (commented):
LGTM but I'd like the same/reverse test for `tests/container`, please

tests/vm
simondeziel: `lxc config get vm1 volatile.base_image` should give you the full fingerprint of the image that's locally cached as it was used as the base image for `vm1`.

kadinsayani:
> LGTM but I'd like the same/reverse test for `tests/container`, please

There is a reverse test included in https://github.com/canonical/lxd/pull/14137. Would you like it replicated in `lxd-ci` as well?

simondeziel:
> > LGTM but I'd like the same/reverse test for `tests/container`, please
> 
> There is a reverse test included in [canonical/lxd#14137](https://github.com/canonical/lxd/pull/14137). Would you like it replicated in `lxd-ci` as well?

Please as this makes the container and VM test "symetric" and also allows checking on multiple snap versions. Also, it's quite cheap in terms of CI time. Thanks!

kadinsayani:
> Please as this makes the container and VM test "symetric" and also allows checking on multiple snap versions. Also, it's quite cheap in terms of CI time. Thanks!

Sounds good - no problem!

simondeziel (approved):

tests/container
```diff
@@ -155,6 +155,13 @@ lxc stop c1
 lxc start c1
 waitInstanceBooted c1
 
+echo "==> Ensure aliased image won't launch with VM flag set"
+lxc stop c1
+lxc image alias create containeralias "$(lxc config get c1 volatile.base_image)"
+! lxc launch containeralias --vm || false
+lxc start c1
+waitInstanceBooted c1
```
simondeziel, line 163: I don't think the alias and `lxc launch` bit needs to be done while `c1` is stopped as the base image is going to be untouched anyway.

tests/vm
```diff
@@ -16,6 +16,10 @@ waitInstanceBooted vm1
 echo "==> Test cleanly stopping a VM"
 lxc stop vm1
 
+echo "==> Ensure that we can't launch an aliased VM image as a container"
+lxc image alias create vmalias "$(lxc config get vm1 volatile.base_image)"
+! lxc launch vmalias || false
```
simondeziel, line 21: As discussed, this breaks the assumption that VM images can be implicitly used to launch VMs without having to provide the `--vm` flag.

simondeziel (approved):

tomponline:
Closing for now as not needed until we decide what we are going to do to resolve the issue.

kadinsayani:
I’ve updated the PR description in case we decide to reopen this at any point.

tests/storage-vm
```diff
@@ -612,7 +612,7 @@ for poolDriver in $poolDriverList; do
         lxc config set storage.images_volume "${poolName}"/images
         lxc publish v1 --alias v1image
         lxc delete v1
-        lxc launch v1image v2 -s "${poolName}"
+        lxc launch v1image v2 --vm -s "${poolName}"
```
simondeziel, line 615: That's the only place where a VM image alias was used to launch an instance without providing that `--vm` flag.

From @tomponline's comment, I understand we don't want to break this now but maybe we should be more explicit in our testing. How about adding a test for that to `tests/vm`?

simondeziel: I'm fine with a comment but I'll submit a PR to add an explicit test for the implicit `--vm` flag ;)

simondeziel: @kadinsayani could you please leave the `--vm` flag out but put a comment as requested by Tom. Thanks

simondeziel (approved):

tomponline (approved):

PR #297, opened 2024-09-26
tests/storage-vm: combine volume.block.filesystem test with volume.size

simondeziel:


tests/storage-vm
```diff
@@ -609,7 +609,7 @@ for poolDriver in $poolDriverList; do
         lxc config set storage.images_volume "${poolName}"/images
         lxc publish v1 --alias v1image
         lxc delete v1
-        lxc launch v1image v2 -s "${poolName}"
+        lxc launch v1image v2 --vm -s "${poolName}"
```
tomponline, line 620: for now we wont want to include this until we decide whether or not we're actually going to make the change, so lets back this out for now.

simondeziel:
While in there, I noticed that all the block based storage drivers were also testing that the `config.mount` disk was using `ext4` rather than `xfs` when the pool was configured to use this later FS. As such, I introduced an equivalent variation for ZFS when used in `volume.zfs.block_mode=true`.

tomponline (approved):

PR #298, opened 2024-10-01
tests/conversion: SELinux is no longer in `images:centos/9-Stream`

simondeziel:
Rather than assume it is there or not, let's only disable it if the config file is present. This should keep it working even if SElinux is reintroduced in the image.


tomponline (approved):

MusicDin (approved):
Thanks

PR #299, opened 2024-10-01
Fix tests/cgroup on cgroup2

mihalicyn:
When cgroup2 is used there is a difference between root of cgroupfs tree and sub-cgroups. When we do a cgroup feature checking we should check for `memory.swap.max` file existence in a `/sys/fs/cgroup/system.slice/memory.swap.max` instead of `/sys/fs/cgroup/memory.swap.max`.

See also:
https://github.com/lxc/lxcfs/commit/f496e62cdbeb01215af34fed266fc5a98d25feeb

### DIFF
--- a/tests/cgroup
+++ b/tests/cgroup
@@ -22,7 +22,7 @@ lxc launch "${IMAGE}" c1
 echo "==> Validate default values"
 [ "$(lxc exec c1 -- nproc)" = "$(nproc)" ]
 [ "$(lxc exec c1 -- grep ^MemTotal /proc/meminfo)" = "$(grep ^MemTotal /proc/meminfo)" ]
-if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/memory.swap.max" ]; then
+if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/system.slice/memory.swap.max" ]; then
     [ "$(lxc exec c1 -- grep ^SwapTotal /proc/meminfo)" = "$(grep ^SwapTotal /proc/meminfo)" ]
 else
     [ "$(lxc exec c1 -- grep ^SwapTotal /proc/meminfo)" = "SwapTotal:             0 kB" ]
@@ -110,7 +110,7 @@ echo "==> Testing memory limits"
 MEM_LIMIT_MIB=512
 lxc config set c1 limits.memory="${MEM_LIMIT_MIB}MiB"
 [ "$(lxc exec c1 -- grep ^MemTotal /proc/meminfo)" = "MemTotal:         $((MEM_LIMIT_MIB * 1024)) kB" ]
-if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/memory.swap.max" ]; then
+if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/system.slice/memory.swap.max" ]; then
     [ "$(lxc exec c1 -- grep ^SwapTotal /proc/meminfo)" = "SwapTotal:        $((MEM_LIMIT_MIB * 1024)) kB" ]
 else
     [ "$(lxc exec c1 -- grep ^SwapTotal /proc/meminfo)" = "SwapTotal:             0 kB" ]
@@ -133,7 +133,7 @@ if [ -e "/sys/fs/cgroup/memory" ]; then
     [ "$(lxc exec c1 -- cat /sys/fs/cgroup/memory/memory.swappiness)" = "65" ]
 fi
 
-if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/memory.swap.max" ]; then
+if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/system.slice/memory.swap.max" ]; then
     lxc config set c1 limits.memory 128MiB
     [ "$(lxc exec c1 -- grep ^MemTotal /proc/meminfo)" = "MemTotal:         131072 kB" ]
 
--- a/tests/cgroup
+++ b/tests/cgroup
@@ -12,6 +12,13 @@ lxd init --auto
 
 IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
 
+rc=0
+journalctl --quiet --no-hostname --no-pager --boot=0 --unit=snap.lxd.daemon.service | grep "Kernel supports swap accounting" || rc="$?"
+LXCFS_SUPPORTS_SWAP_ACCOUNTING=0
+if [ "${rc}" -eq 0 ]; then
+    LXCFS_SUPPORTS_SWAP_ACCOUNTING=1
+fi
+
 # Test
 set -x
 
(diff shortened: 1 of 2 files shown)


mihalicyn:
After merging this we will get test regressions on older LXD versions (when run on cgroup2-enabled hosts). But really those are not regressions and we should update LXCFS to v6.0.2 or backport https://github.com/lxc/lxcfs/pull/617

simondeziel:
`latest/edge` which has the proper LXCFS version is now failing at a later test: https://github.com/canonical/lxd-ci/actions/runs/11128901521/job/30924728690?pr=299#step:8:432

mihalicyn:
> `latest/edge` which has the proper LXCFS version is now failing at a later test: https://github.com/canonical/lxd-ci/actions/runs/11128901521/job/30924728690?pr=299#step:8:432

yeah, I'm on it ;-)

mihalicyn:
Sigh... https://github.com/lxc/lxcfs/pull/661

mihalicyn:
```
==> Testing memory limits
+ MEM_LIMIT_MIB=512
+ lxc config set c1 limits.memory=512MiB
++ lxc exec c1 -- grep '^MemTotal' /proc/meminfo
+ '[' 'MemTotal:       16373812 kB' = 'MemTotal:         524288 kB' ']'
```

on latest/edge - 24.04 will be fixed by https://github.com/lxc/lxcfs/pull/661

```
Test failed
++ grep '^SwapTotal' /proc/meminfo
+ '[' 'SwapTotal:             0 kB' = 'SwapTotal:       4194300 kB' ']'
```

on 5.21/edge - 22.04, 5.0/edge - 22.04, 4.0/edge - 22.04, 5.21/edge - 24.04

TBD

mihalicyn:
I don't think that we should backport https://github.com/lxc/lxcfs/pull/663 to LXD LTS releases (which using LXCFS versions 4.0.X and 5.0.X) as it was never working there. But this fix will be included into the next stable 6.0.X LXCFS release, for sure.

tomponline:
> But this fix will be included into the next stable 6.0.X LXCFS release, for sure.

Can we patch it for latest/candidate and latest/edge so we dont have persistent failures on CI and a regression in the next release?

mihalicyn:
> Can we patch it for latest/candidate and latest/edge so we dont have persistent failures on CI and a regression in the next release?

latest/candidate and latest/edge are using `main` branch of LXCFS so they should be fine.

I have also fixed tests to handle this differences properly.

But it's not the end of a story yet. I need to figure out one more thing which I found weird in there...

mihalicyn:
ah, it's the same thing which I fixed by https://github.com/lxc/lxcfs/pull/661 but now it fails on swap checks. Should be fine. Let's wait until latest/edge is rebuilt.

tests/cgroup
```diff
@@ -12,6 +12,13 @@ lxd init --auto
 
 IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
 
+rc=0
+journalctl --quiet --no-hostname --no-pager --boot=0 --unit=snap.lxd.daemon.service | grep "Kernel supports swap accounting" || rc="$?"
+LXCFS_SUPPORTS_SWAP_ACCOUNTING=0
+if [ "${rc}" -eq 0 ]; then
```
simondeziel, line 18:
```suggestion
LXCFS_SUPPORTS_SWAP_ACCOUNTING=0
if journalctl --quiet --no-hostname --no-pager --boot=0 --unit=snap.lxd.daemon.service --grep "Kernel supports swap accounting"; then
```

Out of curiosity, should my 6.8 (Jammy HWE) kernel meet the needed requirements for this? I don't have that message with a fresh `latest/edge` snap:

```
$ snap list lxd
Name  Version      Rev    Tracking     Publisher   Notes
lxd   git-538cfa3  30507  latest/edge  canonical✓  -
$ uname -a
Linux sdeziel-lemur 6.8.0-45-generic #45~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Sep 11 15:25:05 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
$ cat /proc/cmdline 
BOOT_IMAGE=/vmlinuz-6.8.0-45-generic root=UUID=78a3536d-545e-4ccd-8c89-3e5165da9420 ro rootflags=subvol=@ quiet splash debugfs=off nmi_watchdog=0 nr_cpus=12 panic=300 8250.nr_uarts=1 vsyscall=none vt.handoff=7
```

simondeziel (approved):

tests/cgroup
mihalicyn: It's not really about kernel. This specific check is about LXCFS. The problem is that currently LXCFS is broken on cgroup2 and always says that `Kernel does not support swap accounting` while it is supported by the kernel.

On `latest/edge` you should get LXCFS from the `main` branch where it's fixed by https://github.com/lxc/lxcfs/pull/663

mihalicyn: To conclude, if you have:
- recent LXD from `latest/edge`
- kernel 6.8
- cgroup2

it should say `Kernel supports swap accounting` in lxd daemon logs

simondeziel: OK I tick all the boxes but I had not started LXD after the refresh... my bad. Thanks!

mihalicyn:
@tomponline I think it is good to go now

tests/cgroup
```diff
@@ -22,7 +22,7 @@ lxc launch "${IMAGE}" c1
 echo "==> Validate default values"
 [ "$(lxc exec c1 -- nproc)" = "$(nproc)" ]
 [ "$(lxc exec c1 -- grep ^MemTotal /proc/meminfo)" = "$(grep ^MemTotal /proc/meminfo)" ]
-if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/memory.swap.max" ]; then
+if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/system.slice/memory.swap.max" ]; then
```
tomponline: if this is a LXCFS regression, should we rather than updating our tests to avoid the regression, instead cherry-pick the fix into our snap so that the regression doesn't occur. Because if i'm understanding this correctly, this will give the false impression that the tests are passing, which would then lead us to release a snap with a regressed LXCFS?

mihalicyn: >if this is a LXCFS regression, should we rather than updating our tests to avoid the regression

It is not a LXCFS regression. It was always broken like that with cgroup2.

mihalicyn: https://github.com/lxc/lxcfs/pull/661 fixes a regression, yes. But it only in the `main` branch and only `latest/edge` LXD snap was affected

https://github.com/lxc/lxcfs/pull/663 just makes swap accounting to work properly with cgroup2. It never worked before.

mihalicyn: I started to dive into this stuff because of regression (fixed by https://github.com/lxc/lxcfs/pull/661), but then I discovered that our lxd-ci tests didn't work properly on cgroup2 and after I've fixed this I discovered that LXCFS doesn't handle swap accounting on cgroup2 and fixed this by https://github.com/lxc/lxcfs/pull/663. That's a full explanation, I guess.

mihalicyn: >Because if i'm understanding this correctly, this will give the false impression that the tests are passing, which would then lead us to release a snap with a regressed LXCFS?

no-no, inversely I'm extending our tests to cover cgroup2 scenario. See, we have a checks like:
```
if [ -e "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes" ] || [ -e "/sys/fs/cgroup/memory.swap.max" ]; then
```

first condition is correct (for cgroup v1), second condition is not correct. It will always fail on the host, because `memory.swap.max` does never exist in a root cgroup. Which means that we don't test cgroup2 stuff at all.

tomponline: I see, thanks for the clarification.

tomponline (approved):

PR #300, opened 2024-10-01
Update VM CPU pinning tests to use `limits.cpu.pin_strategy`

kadinsayani:
This PR includes updates to `tests/cpu-vm`.

Changes
- Check if feature is present by checking API extension `limits_cpu_pin_strategy` (landing with https://github.com/canonical/lxd/pull/14171); and
- Launch ephemeral VMs with `limits_cpu_pin_strategy=auto` to check for CPU pinning, and `limits_cpu_pin_strategy=none` to ensure CPU's are not automatically pinned.


tests/cpu-vm
```diff
@@ -15,11 +15,10 @@ if ! hasNeededAPIExtension cpu_hotplug; then
   exit 0
 fi
 
-# required for "CPU auto pinning" feature check
-# as we don't have a separate API extension for it
-# and we rely on the debug output in the LXD daemon logs.
-snap set lxd daemon.debug=true
-systemctl restart snap.lxd.daemon.service
+if ! hasNeededAPIExtension limits_cpu_pin_strategy; then
```
simondeziel, line 18: I wouldn't go as far as skipping the whole test script. Could you rework it so that just the pinning checks depends on that API extension's presence?

tests/cpu-vm
```diff
@@ -86,35 +85,42 @@ done
...
 # Unset CPU limit
 lxc config unset v1 limits.cpu
 
 # Unsetting the limit should leave the VM with 1 CPU
 [ "$(lxc exec v1 -- ls /sys/devices/system/cpu | grep -xEc 'cpu[[:digit:]]+')" -eq "1" ]
 
-echo "==> Stopping and deleting ephemeral VM"
-# Stop VM and check its deleted.
+echo "==> Create ephemeral VM with CPU pinning disabled and boot"
+lxc launch "${IMAGE}" v2 --vm -c limits.cpu.pin_strategy=none -s "${poolName}" --ephemeral
```
simondeziel, line 108: `limits.cpu.pin_strategy=none` being the intended default, I would probably not specify it and simply check that no pinning is applied with default settings. I'd also revert to the old `echo`.

simondeziel: Also, I would try to minimize the number of running/existing instances if possible. As such, I'd keep using `v1` for as long as it make sense and stop + reconfig it if needed/possible. Otherwise, nuking it and creating a fresh `v1` is also possible.

Creating instances has a cost (resource+time) that's higher than reconfiguring an existing one.

simondeziel (commented):
If the API extension is detected, I'd also want a few quick checks trying to set/unset it while the VM is running (expecting failure).

tests/cpu-vm
kadinsayani: Sure thing!

tests/cpu-vm
```diff
@@ -79,16 +73,40 @@ for i in $(seq "${cpuCount}" -1 1); do
...
+  lxc stop -f v1
+  lxc launch "${IMAGE}" v1 --vm -c limits.cpu.pin_strategy=auto -s "${poolName}" --ephemeral
+  waitInstanceReady v1
+  lxc info v1
+
+  lxc config set v1 limits.cpu="${cpuCount}"
+  [ "$(lxc exec v1 -- ls /sys/devices/system/cpu | grep -xEc 'cpu[[:digit:]]+')" -eq "${cpuCount}" ]
+
+  # Try doing pinning while VM is running (shouldn't work)
+  ! lxc config set v1 limits.cpu=1,2 || false
```
simondeziel, line 104: I'd move that back out of the `if` as this doesn't depend on the API's presence.

tests/cpu-vm
```diff
@@ -79,16 +73,40 @@ for i in $(seq "${cpuCount}" -1 1); do
...
+  echo "==> Check CPU auto pinning when limits.cpu.pin_strategy=auto"
+
+  # Stop VM
+  lxc stop -f v1
+  lxc launch "${IMAGE}" v1 --vm -c limits.cpu.pin_strategy=auto -s "${poolName}" --ephemeral
+  waitInstanceReady v1
+  lxc info v1
+
+  lxc config set v1 limits.cpu="${cpuCount}"
+  [ "$(lxc exec v1 -- ls /sys/devices/system/cpu | grep -xEc 'cpu[[:digit:]]+')" -eq "${cpuCount}" ]
```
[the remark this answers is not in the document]
kadinsayani: I've removed the repeated test but I think we still need the `waitInstanceReady` call in order to get the QEMU PID.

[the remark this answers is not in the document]
simondeziel: In theory, QEMU's PID should be immediately available and only `lxc exec` calls would need the `waitInstanceReady`.

tests/cpu-vm
```diff
@@ -79,16 +73,39 @@ for i in $(seq "${cpuCount}" -1 1); do
...
+if hasNeededAPIExtension limits_cpu_pin_strategy; then
+  echo "==> Check CPU auto pinning when limits.cpu.pin_strategy=auto"
+
+  # Stop VM
+  lxc stop -f v1
+  lxc launch "${IMAGE}" v1 --vm -c limits.cpu.pin_strategy=auto -s "${poolName}" --ephemeral
+  waitInstanceReady v1
+  lxc info v1
+
+  lxc config set v1 limits.cpu="${cpuCount}"
```
simondeziel, line 103: If multiple CPUs are desired here, I'd make that a `lxc launch` time config unless hot-plugging is what you want to test.

CPUs available at launch time are put online inside the VM so it makes use of many CPUs. This contrasts with hot-plugged CPUs that are not put `online` once plugged after the boot has completed. In other words, hot-plugged CPUs would remain unused.

tests/cpu-vm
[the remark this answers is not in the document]
kadinsayani: Just checked and you're right, no need for the `waitInstanceReady` call 👍

tests/cpu-vm
kadinsayani: Great suggestion - since we're already testing CPU hot plugging earlier, we can set the number of CPU's at launch time.

kadinsayani: Thanks!

tests/cpu-vm
```diff
@@ -79,16 +73,35 @@ for i in $(seq "${cpuCount}" -1 1); do
   [ "$(lxc exec v1 -- ls /sys/devices/system/cpu | grep -xEc 'cpu[[:digit:]]+')" -eq "${i}" ]
 done
 
-# Try doing pinning while VM is running (shouldn't work)
-! lxc config set v1 limits.cpu=1,2 || false
+echo "==> Check that there is no CPU pinning set"
+QEMU_PID=$(lxc info v1 | awk '/^PID:/ {print $2}')
+! taskset --cpu-list -a -p "${QEMU_PID}" | grep -E ':\s+[0-9]+$' || false
```
simondeziel, line 78: This currently fails on `latest/edge` as it doesn't have the `limits_cpu_pin_strategy` extension but the old autopinning by default.

The autopinning goes away and never landed in a LTS so maybe that's an acceptable situation until https://github.com/canonical/lxd/pull/14171 is merged?

That said, I'd still like a full run on `latest/edge` so maybe stub out that bit for a quick CI run?

kadinsayani: Since the auto pinning will be disabled by default on `latest/edge` with the merge of https://github.com/canonical/lxd/pull/14171, I think we're covered. I've done a full run of `cpu-vm` against `latest/edge` with https://github.com/canonical/lxd/pull/14171's LXD side-loaded and it passed ✅.

simondeziel: Works for me, thanks for confirming!

simondeziel (dismissed):

tomponline (approved):

PR #302, opened 2024-10-03
Add testing for OVN uplink attached to VLAN

boltmark:
This PR adds testing for when an OVN uplink network is attached to a VLAN.

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -2137,6 +2137,45 @@ ovn_leases_tests() {
         lxc network delete lxdbr0
 }
 
+ovn_vlan_uplink_tests() {
+    echo "==> VLAN uplink"
+
+    lxc network create lxdbr0 \
+        ipv4.address=10.10.10.1/24 ipv4.nat=true \
+        ipv4.dhcp.ranges=10.10.10.2-10.10.10.199 \
+
+    echo "==> Create physical UPLINK network with VLAN config set to 10"
+    lxc network create UPLINK --type=physical parent=lxdbr0 \
+        ipv4.ovn.ranges=10.10.10.200-10.10.10.254 \
+        ipv4.gateway=10.10.10.1/24 \
+        vlan=10
+
+    echo "==> Create OVN network"
+    lxc network create ovn-virtual-network --type=ovn network=UPLINK
+    sleep 2
+
+    echo "==> Set up containers"
+    lxc launch "${IMAGE}" c1-ovn -n ovn-virtual-network -s default
+    lxc launch "${IMAGE}" c1-lxdbr0 -n lxdbr0 -s default --device eth0,vlan=10
+    sleep 2
+
+    lxc exec c1-lxdbr0 -- ip addr add 10.10.10.100/24 dev eth0
+    lxc exec c1-lxdbr0 -- ip link set dev eth0 up
+
+    echo "==> Ensure OVN network can reach containers on the configured VLAN"
+    lxc exec c1-ovn -- ping -nc1 -4 -w5 10.10.10.100
+
+    echo "==> Ensure containers on VLAN can reach OVN network"
+    ovn_address=$(lxc network get ovn-virtual-network volatile.network.ipv4.address)
+    lxc exec c1-lxdbr0 -- ping -nc1 -4 -w5 "${ovn_address}"
+
+    echo "==> Cleanup"
+    lxc delete -f c1-lxdbr0 c1-ovn
+    lxc network delete ovn-virtual-network
+    lxc network delete UPLINK
+    lxc network delete lxdbr0
+}
+
 # Allow for running a specific set of tests.
 if [ "$#" -gt 0 ]; then
     TEST_CURRENT="ovn_${1}_tests"
@@ -2162,6 +2201,11 @@ else
       ovn_l3only_tests
     else
       echo "Skipping ovn_l3only_tests, not supported"
+    fi
+    if hasNeededAPIExtension network_ovn_uplink_vlan; then
+      ovn_vlan_uplink_tests
+    else
+      echo "Skipping ovn_vlan_uplink_tests, not supported"
     fi
     ovn_leases_tests
 fi


tests/network-ovn
```diff
@@ -2096,6 +2096,51 @@ ovn_leases_tests() {
...
+
+    echo "==> Create OVN network using a specified VLAN ID"
+    lxc network create ovn-virtual-network --type=ovn network=lxdbr0 vlan=10
+    sleep 2
+
+    echo "==> Ensure localnet port has the tag column set in OVN northbound DB"
+    ovn-nbctl --bare --format=csv --columns=type,tag find logical_switch_port | grep -F "localnet,10"
+
+    echo "==> Ensure OVS configured to accept VLAN traffic"
+    uplink_port=$(ovs-vsctl show | grep "Bridge lxdovn" | awk '{print $2"b"}')
```
simondeziel, line 2122: `grep`'s not strictly needed as `awk` can do string matching too:
```suggestion
    uplink_port=$(ovs-vsctl show | awk '/Bridge lxdovn/ {print $2"b"}')
```

tests/network-ovn
```diff
@@ -2096,6 +2096,51 @@ ovn_leases_tests() {
...
+    ovs-vsctl get port "${uplink_port}" trunks | grep -F "[10]"
+
+    echo "==> Ensure OVN network can reach containers on the configured VLAN"
+    lxc launch "${IMAGE}" c1-ovn -n ovn-virtual-network
+    lxc launch "${IMAGE}" c1-lxdbr0 -n lxdbr0
+    sleep 2
+
+    lxc config device set c1-lxdbr0 eth0 vlan=10
+    lxc exec c1-lxdbr0 -- ip addr add 10.10.10.100/24 dev eth0
+    lxc exec c1-lxdbr0 -- ip link set dev eth0 up
```
simondeziel, line 2133: In my tests, the device remains up even after adding the `vlan=10` config so this should not be needed.

tests/network-ovn
```diff
@@ -2096,6 +2096,51 @@ ovn_leases_tests() {
...
+    lxc exec c1-lxdbr0 -- ip route add default via 10.10.10.1
+    sleep 5
+
+    lxc exec c1-ovn -- ping -nc1 -4 -w5 10.10.10.100
+
+    lxc rm -f c1-lxdbr0 c1-ovn
```
simondeziel, line 2139:
```suggestion
    lxc delete -f c1-lxdbr0 c1-ovn
```

We *try* not to use command aliases.

tests/network-ovn
```diff
@@ -2096,6 +2096,51 @@ ovn_leases_tests() {
...
+    lxc launch "${IMAGE}" c1-lxdbr0 -n lxdbr0
+    sleep 2
+
+    lxc config device set c1-lxdbr0 eth0 vlan=10
+    lxc exec c1-lxdbr0 -- ip addr add 10.10.10.100/24 dev eth0
+    lxc exec c1-lxdbr0 -- ip link set dev eth0 up
+    lxc exec c1-lxdbr0 -- ip route add default via 10.10.10.1
+    sleep 5
+
+    lxc exec c1-ovn -- ping -nc1 -4 -w5 10.10.10.100
```
simondeziel, line 2137: I'd like to see a ping from `c1-lxdbr0` to `c1-ovn`. This will likely require getting `c1-ovn`'s IP with `lxc list -c 4 -f csv c1-ovn | cut -d " " -f1` (or similar) as DNS won't work.

tests/network-ovn
```diff
@@ -2096,6 +2096,51 @@ ovn_leases_tests() {
         lxc network delete lxdbr0
 }
 
+ovn_vlan_uplink_tests() {
```
simondeziel, line 2099: This test will need to be called (at the bottom of the script) if the right API extension is detected. See `hasNeededAPIExtension` in `bin/helpers`.

tests/network-ovn
boltmark: Good to know, thanks! This check has been removed so I'll resolve this.

tests/network-ovn
boltmark: Ah, got it. Fixed!

tests/network-ovn
boltmark: Done!

tests/network-ovn
boltmark: Added

tests/network-ovn
boltmark: If I don't ensure `eth0` is up then I see the next command to add the default route fail:
```
$ lxc exec c1-lxdbr0 -- ip route add default via 10.10.10.1
Error: Nexthop has invalid gateway.
```

Am I missing something here?

tests/network-ovn
```diff
@@ -2127,6 +2127,55 @@ ovn_leases_tests() {
...
+        ipv4.gateway=10.10.10.1/24 \
+        vlan=10
+
+    echo "==> Create OVN network"
+    lxc network create ovn-virtual-network --type=ovn network=UPLINK
+    sleep 2
+
+    echo "==> Set up containers"
+    lxc launch ubuntu:24.04 c1-ovn -n ovn-virtual-network
+    lxc launch ubuntu:24.04 c1-lxdbr0 -n lxdbr0
```
simondeziel, line 2156: Please use `"${IMAGE}"` here.

tests/network-ovn
```diff
@@ -2127,6 +2127,55 @@ ovn_leases_tests() {
...
+    bridge vlan add dev lxdbr0 vid 10 tagged self
```
simondeziel, line 2157: Does that allow VLAN tagged packets/frames to be accepted by `lxdbr0` rather than being discarded? If so, a comment to that effect would help (me at least ;) ).

tests/network-ovn
simondeziel: Hmm, sounds like a race as the `up` part is not needed locally but I've found some other occurrences https://serverfault.com/questions/953198/why-nexthop-has-invalid-gateway-when-it-seems-to-be-defined

It doesn't hurt so please keep it and thanks for checking. TIL

tests/network-ovn
boltmark: Fixed!

tests/network-ovn
```diff
@@ -2127,6 +2127,57 @@ ovn_leases_tests() {
         lxc network delete lxdbr0
 }
 
+ovn_vlan_uplink_tests() {
+    echo "==> vlan uplink"
+
+    sudo ovs-vsctl set open_vswitch . \
```
tomponline, line 2133: this isnt needed as its done at the top of the script.

tests/network-ovn
```diff
@@ -2127,6 +2127,57 @@ ovn_leases_tests() {
...
+        external_ids:ovn-remote=unix:/var/run/ovn/ovnsb_db.sock \
+        external_ids:ovn-encap-type=geneve \
+        external_ids:ovn-encap-ip=127.0.0.1
+
+    lxc network create lxdbr0 \
+        ipv4.address=10.10.10.1/24 ipv4.nat=true \
+        ipv4.dhcp.ranges=10.10.10.2-10.10.10.199 \
+        ipv6.address=fd42:4242:4242:1010::1/64 ipv6.nat=true \
+        ipv6.ovn.ranges=fd42:4242:4242:1010::200-fd42:4242:4242:1010::254
```
tomponline, line 2142: why using ipv6.ovn.ranges on the lxdbr0 network?

tests/network-ovn
```diff
@@ -2127,6 +2127,57 @@ ovn_leases_tests() {
...
+        ipv4.gateway=10.10.10.1/24 \
+        vlan=10
+
+    echo "==> Create OVN network"
+    lxc network create ovn-virtual-network --type=ovn network=UPLINK
+    sleep 2
+
+    echo "==> Set up containers"
+    lxc launch "${IMAGE}" c1-ovn -n ovn-virtual-network -s default
+    lxc launch "${IMAGE}" c1-lxdbr0 -n lxdbr0 -s default
```
tomponline, line 2156: You can avoid the need for the `lxc config device set ...` below by adding here `--device eth0,vlan=10`

tests/network-ovn
```diff
@@ -2127,6 +2127,57 @@ ovn_leases_tests() {
...
+    sleep 2
+
+    # Configure lxdbr0 to handle tagged traffic for VLAN 10 and retain the VLAN tag
+    bridge vlan add dev lxdbr0 vid 10 tagged self
```
tomponline, line 2160: this shouldnt be needed

tests/network-ovn
```diff
@@ -2154,6 +2205,7 @@ else
       echo "Skipping ovn_l3only_tests, not supported"
     fi
     ovn_leases_tests
+    ovn_vlan_uplink_tests
```
tomponline, line 2208: @boltmark did this test work before https://github.com/canonical/lxd/pull/14196 was merged? 

If not then you're going to need to add an API extension please so that the test can detect support for using a bridge network as an uplink with VLAN.

Suggest `network_ovn_uplink_vlan`

tests/network-ovn
```diff
@@ -2127,6 +2127,57 @@ ovn_leases_tests() {
         lxc network delete lxdbr0
 }
 
+ovn_vlan_uplink_tests() {
+    echo "==> vlan uplink"
```
tomponline, line 2131: Please change "vlan" to "VLAN" in end user messages as its an acronym.

boltmark: Done

tests/network-ovn
boltmark: Removed

tests/network-ovn
boltmark: Removed

tests/network-ovn
boltmark: Ah, thanks! Fixed

tests/network-ovn
```diff
@@ -2127,6 +2127,57 @@ ovn_leases_tests() {
...
+    lxc launch "${IMAGE}" c1-lxdbr0 -n lxdbr0 -s default
+    sleep 2
+
+    # Configure lxdbr0 to handle tagged traffic for VLAN 10 and retain the VLAN tag
+    bridge vlan add dev lxdbr0 vid 10 tagged self
+
+    lxc config device set c1-lxdbr0 eth0 vlan=10
+    lxc exec c1-lxdbr0 -- ip addr add 10.10.10.100/24 dev eth0
+    lxc exec c1-lxdbr0 -- ip link set dev eth0 up
+    lxc exec c1-lxdbr0 -- ip route add default via 10.10.10.1
```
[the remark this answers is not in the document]
boltmark: Thanks! Removed

tests/network-ovn
boltmark: Removed, thanks

tests/network-ovn
boltmark: Hi @tomponline, the test passes on older lxd versions if we keep

`bridge vlan add dev lxdbr0 vid 10 tagged self`

When it is removed, the test only passes on `latest/edge`. I suppose we should add the API extension then?

tomponline: Yes exactly, thats why we need an API extension check, otherwise you're not testing the functionality in LXD instead the test is just setting up the environment for LXD which defeats the purpose of the test :)

boltmark: Got it, thanks! PR for this here: https://github.com/canonical/lxd/pull/14296.

tomponline (approved):

PR #303, opened 2024-10-03
images/kali: Fix enabling cloud-init service

MusicDin:
When cloud-init is installed, the `systemctl enable cloud-init` fails. The service is now called `cloud-init-main`.

Passing build: https://github.com/MusicDin/lxd-ci/actions/runs/11165385223

### DIFF
--- a/images/kali.yaml
+++ b/images/kali.yaml
@@ -1774,7 +1774,7 @@ actions:
     set -eux
 
     # Enable cloud-init units
-    systemctl enable cloud-init
+    systemctl enable cloud-init-main
   variants:
   - cloud
 


images/kali.yaml
```diff
@@ -1774,7 +1774,7 @@ actions:
     set -eux
 
     # Enable cloud-init units
-    systemctl enable cloud-init
+    systemctl enable cloud-init-main
```
simondeziel, line 1777: That's also the case on Ubuntu 24.10 BTW so we might eventually need to do `systemctl enable cloud-init || systemctl enable cloud-init-main` for every images.

simondeziel (approved):

PR #306, opened 2024-10-04
Date handling on weekly news

hamistao:
I was messing with this the other day and thought this could be useful :D

This does two things:
- Fills the dates on the output of `weekly-changes.sh`
- If the second date is not provided, assume a Monday to Sunday interval (6 days after the first date)


simondeziel (approved):

MusicDin:
Thanks, very useful addition 👍

PR #310, opened 2024-10-08
tests/storage-vm: Add test for live resizing of VM disks with zfs or lvm pool driver

kadinsayani:
This PR adds tests for https://github.com/canonical/lxd/pull/14211. The test depends on presence of the feature which is indicated by the api extension `vm_storage_disk_live_resize`.

### DIFF
--- a/tests/storage-vm
+++ b/tests/storage-vm
@@ -281,31 +281,49 @@ for poolDriver in $poolDriverList; do
         echo "==> Testing VM optimized export/import (while running to check config.mount is excluded)"
         ! lxc export v1 --quiet --compression=none - --optimized-storage | tar -tf - | grep -F config.mount || false
 
-        echo "==> Increasing VM root disk size for next boot (24GiB)"
-        lxc config device set v1 root size=24GiB
-        [ "$(lxc config get v1 volatile.root.apply_quota)" = "true" ]
-
-        if [ "${poolDriver}" != "powerflex" ]; then
-                echo "==> Check VM online shrink is allowed if it greater or equal to initial size"
-                lxc config device set v1 root size=3584MiB # initial
-                lxc config device set v1 root size=32GiB
-
-                echo "==> Setting VM root disk size for next boot (4GiB)"
-                lxc config device set v1 root size=4GiB # final
-        else
-                echo "==> Check VM online shrink is allowed if it greater or equal to initial size"
-                lxc config device set v1 root size=8GiB # initial
-                lxc config device set v1 root size=32GiB
-
-                echo "==> Setting VM root disk size for next boot (16GiB)"
-                lxc config device set v1 root size=16GiB # final
+        LIVE_GROWING_SUPPORTED=0
+        if hasNeededAPIExtension vm_zfs_storage_disk_live_resize && [ "${poolDriver}" = "zfs" ]; then
+                LIVE_GROWING_SUPPORTED=1
+        elif hasNeededAPIExtension vm_lvm_storage_disk_live_resize; then
+                if [ "${poolDriver}" = "lvm" ] || [ "${poolDriver}" = "lvm-thin" ]; then
+                        LIVE_GROWING_SUPPORTED=1
+                fi
         fi
-        [ "$(lxc config get v1 volatile.root.apply_quota)" = "true" ]
-        lxc stop -f v1
 
-        # Proceed to actually growing the root disk
-        lxc start v1
-        [ "$(lxc config get v1 volatile.root.apply_quota)" = "" ]
+        if [ "${LIVE_GROWING_SUPPORTED}" = 1 ]; then
+                echo "==> Checking live root disk device resize support with ${poolDriver} backend"
+                lxc config device set v1 root size=4GiB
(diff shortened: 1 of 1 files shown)


tests/storage-vm
```diff
@@ -281,6 +281,12 @@ for poolDriver in $poolDriverList; do
         echo "==> Testing VM optimized export/import (while running to check config.mount is excluded)"
         ! lxc export v1 --quiet --compression=none - --optimized-storage | tar -tf - | grep -F config.mount || false
 
+        if [[ $(hasNeededAPIExtension vm_zfs_storage_disk_live_resize) && $(hasNeededAPIExtension vm_lvm_storage_disk_live_resize) && ( "${poolDriver}" == "zfs" || "${poolDriver}" == "lvm" ) ]]; then
```
simondeziel, line 284: I'd simplify the gating logic to handle only 1 extension + driver combo. Also, I think that `lvm-thin` would also work?

```suggestion
        LIVE_GROWING_SUPPORTED=0
        if hasNeededAPIExtension vm_zfs_storage_disk_live_resize && [ "${poolDriver}" = "zfs" ]; then
          LIVE_GROWING_SUPPORTED=1
        elif hasNeededAPIExtension vm_lvm_storage_disk_live_resize; then
          if [ "${poolDriver}" = "lvm" ] || [ "${poolDriver}" = "lvm-thin" ]; then
            LIVE_GROWING_SUPPORTED=1
          fi
        fi
```
The `$()` around `hasNeededAPIExtension vm_zfs_storage_disk_live_resize` is not needed as we only care about the function's return code, not its output.

tests/storage-vm
```diff
@@ -281,6 +281,12 @@ for poolDriver in $poolDriverList; do
...
+                echo "==> Checking live root disk device resize support with ${poolDriver} backend"
+                lxc config device set v1 root size=16GiB
+                [ "$(($(lxc exec v1 -- blockdev --getsize64 /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root) / GiB))" -eq "16" ]
```
simondeziel, line 287: At that point, I'd also want to double check no resize is scheduled/deferred to next boot.

```
[ "$(lxc config device get v1 volatile.root.apply_quota)" = "" ]
```

tests/storage-vm
```diff
@@ -281,6 +281,12 @@ for poolDriver in $poolDriverList; do
...
+        fi
+
         echo "==> Increasing VM root disk size for next boot (24GiB)"
         lxc config device set v1 root size=24GiB
         [ "$(lxc config get v1 volatile.root.apply_quota)" = "true" ]
```
simondeziel, line 292: I suspect this ^ will fail on the live-growable backends as the growing won't be deferred anymore.

simondeziel (changes requested):

tests/storage-vm
kadinsayani: This should pass with new change - here's a copy of my local test with a build of https://github.com/canonical/lxd/pull/14211. 

```bash
❯ lxc launch ubuntu:n --vm v1 --storage=zfs
Launching v1
                                                   
❯ lxc config device set v1 root size=16GiB

❯ lxc exec v1 -- blockdev --getsize64 /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root | awk '{printf "%.2f GiB\n", $1 / (1024^3)}'
```

tests/storage-vm
kadinsayani: Great suggestion - thanks! And yes, `lvm-thin` should also work.

tests/storage-vm
```diff
@@ -281,9 +281,25 @@ for poolDriver in $poolDriverList; do
...
+                [ "$(($(lxc exec v1 -- blockdev --getsize64 /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root) / GiB))" -eq "16" ]
+                [ "$(lxc config get v1 volatile.root.apply_quota)" = "" ]
+        else
+                echo "==> Increasing VM root disk size for next boot (24GiB)"
+                lxc config device set v1 root size=24GiB
+                [ "$(lxc config get v1 volatile.root.apply_quota)" = "true" ]
+        fi
 
         if [ "${poolDriver}" != "powerflex" ]; then
                 echo "==> Check VM online shrink is allowed if it greater or equal to initial size"
```
simondeziel, line 305: Does that still holds if the resize was applied live? I don't think we can go from `small` -> `large` -> `medium` because that involves actually shrinking rather than playing with an uncommitted variable (the future size/quota to apply).

kadinsayani: I've changed the flow of execution to resolve this - in either case the root disk size is 4GiB after the resize tests.

simondeziel (approved):
If all storage drivers are to grow support for live resize then the extension name will change before landing in LXD.

As for the testing logic, it looks good to me even if I've not seen it in action for the `LIVE_GROWING_SUPPORTED=1` case.

kadinsayani:
Closing until work is resumed on https://github.com/canonical/lxd/pull/14211.

PR #311, opened 2024-10-08
Microceph

simondeziel:


simondeziel:
@tomponline as green as it gets. I included more than the `setup-microceph` action integration as I've been sitting on that one for long enough to have many retries :)

I've kept the `tests/vm-migration` using the older `microceph` integration as I will update it once I'm done with making that `setup-microceph` action consumable locally.

tests/efi-vars-editor-vm
```diff
@@ -33,8 +33,6 @@ lxc storage create "${poolName}" "${poolDriver}"
 
 echo "==> Create VM and boot"
 lxc launch "${IMAGE}" v1 --vm -s "${poolName}"
-waitInstanceReady v1
```
tomponline, line 36: This needs to stay (although with a comment would help) as otherwise lxd tries to access the nvram whilst qemu guest initializing it (git history may show more details)

tomponline (changes requested):
All good except the uefi vars editor change

tests/efi-vars-editor-vm
simondeziel: Indeed, found commit a9d485147ca1cdc04 and added a comment, thanks!

simondeziel:
@tomponline all green now and the `efi-vars-editor-vm` change was replaced by a comment, as suggested.

tomponline (approved):

PR #315, opened 2024-10-09
github: workflow reorg

simondeziel:


tomponline (dismissed):

simondeziel:
@tomponline I'm now wondering if we should skip running the log tests once PRs are merged. That feels unwarranted and and burns like 49 hours of CI.

tomponline:
> @tomponline I'm now wondering if we should skip running the tests job once PRs are merged. That feels unwarranted and and burns like 49 hours of CI.

yeah makes sense

tomponline (approved):

PR #319, opened 2024-10-11
Improve testing a stop in the middle of lxc exec

hamistao:
When cleanly stopping a VM in the middle of lxc exec, two things can happen:
 1- The agent is stopped first, in which case the connection with LXD is severed and we are supposed to get a 129 exit code on the client;
 2- The spawned process is stopped first and the agent is able to respond before it also stops. In this case, we are supposed to get the exit code based on the signal that was used to stop the process;
 
 This adds tests for these two scenarios separately.

### DIFF
--- a/tests/vm
+++ b/tests/vm
@@ -21,15 +21,20 @@ if ! echo "${LXD_SNAP_CHANNEL}" | grep -qE '^4\.0/'; then
     lxc start vm1
     waitInstanceBooted vm1
 
-    echo "==> Test lxc exec exit code when stopping the VM cleanly"
-    (sleep 5 && lxc stop vm1) &
-    lxc exec vm1 -- sleep 60 || exitCode=$?
-    [ "${exitCode:-0}" -eq 129 ]
-    wait $!
+    echo "==> Test getting the correct exit code for a signaled process"
+    # Signaling the process spawned by lxc exec and checking its exit code.
+    # Simulates what can happen if the container stops in the middle of lxc exec.
+    for signal in SIGTERM SIGHUP SIGTERM; do
+        echo "==> Test getting the correct exit code when sending ${signal} to an exec'ed process"
+        code="$(kill -l "${signal}")"
+        # killall is not included in minimal images so we use escaping to combine pgrep and kill.
+        (sleep 5 && lxc exec vm1 -- bash -c "kill -s ${signal} \$(pgrep sleep)") &
+        lxc exec vm1 -- sleep 60 || exitCode=$?
+        # Check exit code, should be 128 plus signal number, 1 for SIGHUP, 9 for SIGKILL and 15 for SIGTERM.
+        [ "${exitCode:-0}" -eq $((128 + code)) ]
+    done
 
     echo "==> Test lxc exec exit code when stopping the VM abruptly"
-    lxc start vm1
-    waitInstanceBooted vm1
     (sleep 5 && lxc stop -f vm1) &
     lxc exec vm1 -- sleep 60 || exitCode=$?
     [ "${exitCode:-0}" -eq 129 ]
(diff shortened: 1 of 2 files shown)


simondeziel (changes requested):
`s/x1/vm1/` ?

tests/vm
```diff
@@ -34,6 +43,12 @@ if ! echo "${LXD_SNAP_CHANNEL}" | grep -qE '^4\.0/'; then
     lxc exec vm1 -- sleep 60 || exitCode=$?
     [ "${exitCode:-0}" -eq 129 ]
     wait $!
+
+    echo "==> Test what happens when the connection with the agent is severed"
+    # Simulates what happens during lxc stop
+    lxc exec vm1 -- killall -s SIGTERM lxd-agent || exitCode=$?
+    [ "${exitCode:-0}" -eq 129 ]
+    wait $!
```
simondeziel, line 51: I don't think there is any background job to wait for at this point, is there?

tests/vm
```diff
@@ -21,11 +21,20 @@ if ! echo "${LXD_SNAP_CHANNEL}" | grep -qE '^4\.0/'; then
     lxc start vm1
     waitInstanceBooted vm1
 
-    echo "==> Test lxc exec exit code when stopping the VM cleanly"
-    (sleep 5 && lxc stop vm1) &
+    echo "==> Test getting the correct exit code for a signaled process"
+    # Signaling the process spawned by lxc exec and checking its exit code.
+    # Simulates what can happen if the container stops in the middle of lxc exec.
+    (sleep 5 && lxc exec vm1 -- killall -s SIGTERM sleep) &
```
simondeziel, line 27: Could we have a for loop à la:

```
for signal in SIGTERM SIGHUP SIGTERM; do
  echo "==> Test getting the correct exit code when sending ${signal} to an exec'ed process"
  code="$(kill -l "${signal}")
  (sleep 5 && lxc exec vm1 -- killall -s "${signal}" sleep) &
  lxc exec vm1 -- sleep 60 || exitCode=$?
  [ "${exitCode:-0}" -eq $((128 + code)) ]
done 
```

simondeziel (commented):
LGTM with a small proposed twist. Also, at the end of this PR, please keep `tests/container` "in sync" for the interrupted `lxc exec` tests, please.

tests/vm
hamistao: good idea, thanks!

tests/vm
hamistao: Whoops, that slipped me, thanks for catching

hamistao:
@simondeziel @tomponline All passing now! Ran it a bunch of times to be sure and seems fine.

tests/vm
```diff
@@ -21,15 +21,20 @@ if ! echo "${LXD_SNAP_CHANNEL}" | grep -qE '^4\.0/'; then
...
     waitInstanceBooted vm1
 
-    echo "==> Test lxc exec exit code when stopping the VM cleanly"
-    (sleep 5 && lxc stop vm1) &
-    lxc exec vm1 -- sleep 60 || exitCode=$?
-    [ "${exitCode:-0}" -eq 129 ]
-    wait $!
+    echo "==> Test getting the correct exit code for a signaled process"
+    # Signaling the process spawned by lxc exec and checking its exit code.
+    # Simulates what can happen if the container stops in the middle of lxc exec.
```
tomponline, line 26: "container" in the `tests/vm` suite?

tomponline (approved):
Apart from the copy paste issue this looks good

tests/vm
hamistao: whoops, sorry about that. My local version is correct but I forgot to push this change :woozy_face:

PR #322, opened 2024-10-15
Firewall: Add NAT checks

roosterfish:
See https://github.com/canonical/lxd/pull/14279

### DIFF
--- a/tests/network-bridge-firewall
+++ b/tests/network-bridge-firewall
@@ -117,6 +117,83 @@ firewallTests() {
     ! lxc exec c1 -- ping -c1 2001:db8::1 || false
 
     lxc delete -f c1
+
+    # Perform NAT tests for managed bridges only.
+    if [ $managed -eq 1 ]; then
+        echo "=> Performing NAT tests with masquerading"
+        lxc network set lxdbr0 ipv4.nat true
+        lxc network set lxdbr0 ipv6.nat true
+        lxc network create lxdbr1 \
+            ipv4.address=192.0.3.1/24 \
+            ipv6.address=2001:db9::1/64 \
+            ipv4.dhcp.ranges=192.0.3.2-192.0.3.199 \
+            ipv6.dhcp.stateful=true \
+            ipv4.nat=true \
+            ipv6.nat=true
+
+        lxc launch "${IMAGE}" c1
+        lxc launch "${IMAGE}" c2 -n lxdbr1
+        sleep 10
+
+        lxc exec c2 -- ip a add 192.0.3.254/24 dev eth0
+        lxc exec c2 -- ip a add 2001:db9::254/64 dev eth0 nodad
+
+        # Check if regular traffic passes between different networks and is masqueraded with the gateways IP.
+        lxc exec c1 -T -n -- ping -n -4 -w5 192.0.3.254 || true &
+        timeout 5s tcpdump -i lxdbr1 -nn icmp and src 192.0.3.1 -q -c 1 > /dev/null
+        wait
+        lxc exec c1 -T -n -- ping -n -6 -w5 2001:db9::254 || true &
+        timeout 5s tcpdump -i lxdbr1 -nn icmp6 and src 2001:db9::1 -q -c 1 > /dev/null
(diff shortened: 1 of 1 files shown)


tests/network-bridge-firewall
```diff
@@ -116,6 +116,35 @@ firewallTests() {
...
+        ipv4.address=192.0.3.1/24 \
+        ipv6.address=2001:db9::1/64 \
+
+    lxc exec c1 -- ip a add 192.0.2.254/24 dev eth0
+    lxc exec c1 -- ip r replace default via 192.0.2.1 dev eth0
+    lxc exec c1 -- ip a add 2001:db8::254/64 dev eth0 nodad
+    lxc exec c1 -- ip -6 r replace default via 2001:db8::1 dev eth0
+
+    lxc exec c1 -- ping -c1 192.0.3.1
+    lxc exec c1 -- ping -c1 2001:db9::1
```
simondeziel, line 138: Should we also also validate that `tcpdump` with the appropriate filters catches those packets? This would confirm we have the expected source address as mangled by the NAT.

tests/network-bridge-firewall
```diff
@@ -116,6 +116,35 @@ firewallTests() {
...
+
+    echo "=> Performing NAT tests with explicit SNAT address"
+    lxc network set lxdbr0 ipv4.nat.address 192.0.2.10
+    lxc network set lxdbr0 ipv6.nat.address 2001:db8::10
+
+    lxc exec c1 -- ping -c1 192.0.3.1
+    lxc exec c1 -- ping -c1 2001:db9::1
```
simondeziel, line 145: Same suggestion about `tcpdump` here. `tests/network-ovn` have such capture examples if needed.

simondeziel (approved):
LGTM but I think we could add extra validation on the resulting packets.

tests/network-bridge-firewall
roosterfish: Thanks for the pointer I'll have a look. 

Also I am trying to have some multicast tests to properly validate the change in LXD.

tests/network-bridge-firewall
roosterfish: Fixed. I have now added the tcpdump tests for both default masquerading and SNAT IP.

tests/network-bridge-firewall
roosterfish: Fixed.

roosterfish:
Something interesting I have witnessed is if you have the following topology, the default masquerading (withing setting an explicit SNAT IP) uses the IP of the second bridge as source IP. Probably related to the fact that both are located on the same Linux host:

```
instance1 ---> lxdbr0 ---> lxdbr1 ---> instance2
```

In the example above if you ping `instance2` from `instance1` and `ipv4.nat=true` on `lxdbr0` then if you `tcpdump -i lxdbr1 icmp` you see the address of `lxdbr1` as the source address and not the one of `lxdbr0`. But if you set `ipv4.nat.address` on `lxdbr0` you see it as the source address when running `tcpdump -i lxdbr1 icmp`.

roosterfish:
And the [PR on LXD](https://github.com/canonical/lxd/pull/14279) from yesterday fixed multicast as currently only the LXD `latest/edge` test is succeeding https://github.com/canonical/lxd-ci/actions/runs/11366247597/job/31616276114?pr=322.

tests/network-bridge-firewall
```diff
@@ -117,6 +117,71 @@ firewallTests() {
     ! lxc exec c1 -- ping -c1 2001:db8::1 || false
 
     lxc delete -f c1
+
+    # Perform NAT tests for managed bridges only.
+    if [ $managed -eq 1 ]; then
```
tomponline, line 122: we're going to want a condition for testing this only on lxd 5.21 onwards as i dont intend to backport the fix back to 5.0

roosterfish: Fixed. Now the only failing tests are the ones for `5.21/edge` as it's not yet backported.

tomponline (approved):

PR #324, opened 2024-10-22
Refresh or install LXD snap instead of purge+install

simondeziel:


README.md
```diff
@@ -6,13 +6,6 @@ To run a test locally (directly where you invoke it), use the `bin/local-run` he
 ./bin/local-run tests/interception latest/edge
 ```
 
-For faster repeated runs, you might want to tell `snap` that it can purge the LXD snap
-without taking any snapshot:
-
-```
-PURGE_LXD=1 ./bin/local-run tests/interception latest/edge
```
tomponline, line 13: We can keep the PURGE_LXD feature and this help text, we just dont need to use it for the Github runners.

tomponline (approved):
Good idea to reuse!

PR #327, opened 2024-10-24
github: abort system-tests after 60 minutes

simondeziel:
Out of the blue, `tests/vm-nesting` became our biggest [CI consumer](https://github.com/canonical/lxd-ci/actions/metrics/usage?tab=jobs). This was because at some point, that test simply stopped making progress so instead of taking < 15 minutes it would hang for 6 hours and eventually be [cancelled due to timeout](https://github.com/canonical/lxd-ci/actions/runs/11282145005/job/31379018916#step:9:3258).

Our current slowest test is `tests/storage-lvm` when ran on `24.04` and even this very slow combo finishes inside ~45 minutes. Allowing up to twice as long before timing out should provide enough headroom while also limiting the CI waste if a similar hang ever appears again.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -73,6 +73,8 @@ jobs:
   system-tests:
     name: ${{ matrix.test }} (${{ matrix.track }} - ${{ matrix.os }})
     runs-on: ubuntu-${{ matrix.os }}
+    # avoid runaway test burning 6 hours of CI
+    timeout-minutes: 60
     permissions:
       # need that to manipulate caches
       actions: write


simondeziel:
Maybe 90 minutes is even too generous and we should go with 60 minutes and adjust as needed when we hit the limit? @tomponline, WDYT?

tomponline (dismissed):

tomponline (approved):

tomponline:
> Our current slowest test is `tests/storage-lvm` when ran on `24.04`

This certainly warrants more investigation, if lvm is slower than ceph then something is terribly wrong, especially if specific to ubuntu 24.04.

simondeziel:
> > Our current slowest test is `tests/storage-lvm` when ran on `24.04`
> 
> This certainly warrants more investigation, if lvm is slower than ceph then something is terribly wrong, especially if specific to ubuntu 24.04.

Yep, it worries me as well and I intend to dig deeper into this as time permits.

simondeziel:
> > This certainly warrants more investigation, if lvm is slower than ceph then something is terribly wrong, especially if specific to ubuntu 24.04.
> 
> Yep, it worries me as well and I intend to dig deeper into this as time permits.

I suspect it's nothing to do with Ubuntu 24.04 but more with LXD itself, https://github.com/canonical/lxd/issues/14341 contains my analysis.

PR #329, opened 2024-10-25
tests/storage-vm: replace snap restart by systemctl restart

simondeziel:
Workaround for https://bugs.launchpad.net/snapd/+bug/2028141 and https://bugs.launchpad.net/snapd/+bug/2083961

### DIFF
--- a/tests/storage-vm
+++ b/tests/storage-vm
@@ -24,7 +24,7 @@ GiB="$((1024 * MiB))"
 if echo "${poolDriverList}" | grep -qwF "zfs" && grep -qF zfs.external /snap/lxd/current/commands/daemon.start; then
     echo "==> Use external ZFS tools"
     snap set lxd zfs.external=true
-    snap restart lxd
+    systemctl restart snap.lxd.daemon.service
 
     if command -v zfs; then
         apt-get autopurge -y zfsutils-linux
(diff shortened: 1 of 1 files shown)


tests/storage-vm
```diff
@@ -21,7 +21,7 @@ MiB="$((1024 * 1024))"
 GiB="$((1024 * MiB))"
 
 # zfs.external
-if echo "${poolDriverList}" | grep -qwF "zfs" && grep -qF zfs.external /snap/lxd/current/commands/daemon.start; then
+if echo "${poolDriverList}" | grep -qwF "zfs" && grep -qwFm1 zfs_external /snap/lxd/current/commands/daemon.start; then
```
tomponline, line 24: please can you add the *why* to this commit message

simondeziel: It's mostly about checking for `zfs_external` instead of `zfs.external` as mentioned in the commit message. The "dot variant" being present in an log message:

```
# /snap/lxd/current/commands/daemon.start
...
     elif [ -n "${ZFS_VER}" ]; then
        echo "==> Unsupported ZFS version (${ZFS_VER})"
        echo "Consider installing ZFS tools in the host and use zfs.external"
```

I figured it was safer to rely on the actual config name which is the underscore variant.

The `grep` option changes are just me a bit obsessive ;) The `-w` bit meant it will match for whole words and the `-m1` means to stop searching after the 1st match. Here it is in action without `-q` to make it clear:

```
$ grep -wF zfs_external /snap/lxd/current/commands/daemon.start
if [ "${zfs_external:-"false"}" = "true" ]; then
if [ "${zfs_external:-"false"}" = "false" ]; then

$ grep -wFm1 zfs_external /snap/lxd/current/commands/daemon.start
if [ "${zfs_external:-"false"}" = "true" ]; then
```

Is any of this worth adding to the commit message?

tomponline: Thanks. So this is checking if the snap supports zfs.external - this is what i was missing

tomponline (approved):

PR #332, opened 2024-10-25
tests/storage-vm: discard v1/snap0 as soon as unneeded

simondeziel:


simondeziel:
@tomponline with this tweak, the `tests/storage-vm lvm` slow test goes from ~44 minutes to ~37 minutes.

tests/storage-vm
```diff
@@ -238,22 +239,21 @@ for poolDriver in $poolDriverList; do
                 echo "==> Checking VM snapshot copy root disk size is 8GiB"
                 [ "$(($(lxc exec v2 -- blockdev --getsize64 /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_root) / GiB))" -eq "8" ]
         fi
-        lxc delete -f v2
 
         echo "==> Checking moving VM to a different storage pool"
-        lxc copy v1 v2 --stateless
```
tomponline, line 244: Assuming ive read this right, overall this seems to be testing fewer scenarios:

1. Copying a running VM with snapshots via stateless mode.
2. Moving a VM with snapshots across storage pools.

So whilst it is no doubt quicker, i feel it would be better to test the more common scenarios of moving a VM and its snapshots.

simondeziel: Regarding 1), stateless VM copies are done earlier:

```
        echo "==> Checking VM can be migrated with snapshots (same storage pool)"
        lxc copy v1 localhost:v2 --stateless
        [ "$(lxc query /1.0/instances/v2?recursion=1 | jq '.snapshots | length')" -eq "1" ]
        lxc start v2
        waitInstanceReady v2
        lxc stop -f v2

...

        echo "==> Checking VM can be migrated with snapshots (different storage pool)"
        lxc copy v1 localhost:v2 -s "${poolName}2" --stateless
        [ "$(lxc query /1.0/instances/v2?recursion=1 | jq '.snapshots | length')" -eq "1" ]
        lxc start v2
        waitInstanceReady v2
        lxc stop -f v2
```

Granted that's slightly different than the one I moved as `localhost:` is used rather than using the default `local:` remote. Is that a significant enough difference worth reintroducing it?

Regarding 2), I flipped the move directions around but there are still 2 moves of `v2` between pools. The commit message captures it, possibly not clearly enough though?

tomponline: lets chat about this in our next 1:1

simondeziel: Added to our meeting notes (I have other items piling up ;) )

simondeziel:
@tomponline all green except the brittle network-ovn that is unrelated.

tomponline:
Thanks.

In this case the 5.21 ovn tests are failing as ive not backported a bug fix yet.

tomponline (approved):

PR #333, opened 2024-10-26
tests: Ensure network forward create fails with other OVN volatile IP

boltmark:
This PR adds a quick check to ensure that network forward creates fail when referencing the volatile IP of another OVN network on the same uplink.

Draft for now as the lxd PR is not merged.


simondeziel (commented):
Tests are failing, does that depend on a fix/PR yet to be merged in LXD?

boltmark:
> Tests are failing, does that depend on a fix/PR yet to be merged in LXD?

The last run was before the merge of the fix (https://github.com/canonical/lxd/pull/14344), but I have just rebased and ran everything again. Should succeed now for `network-ovn `on `latest/edge`, and I think we are backporting this so that it will succeed on older versions as well.

tomponline (approved):
Ta

PR #334, opened 2024-10-28
images/ubuntu: Enable cloud-init

MusicDin:


simondeziel (approved):

tomponline (commented):
Were we building cloud variants before?

MusicDin:
The desktop image already has cloud-init installed, it was just missing the templates. However, having cloud-init under installed packages will ensure it is always installed (if ever removed) and updated if already installed.

MusicDin:
> Were we building cloud variants before?

Yes, we were building cloud images, but for non-desktop variants. Therefore, cloud-init part was a leftover that we have simply reused for desktop images.

PR #335, opened 2024-10-28
bin/test-image: check if cloud-init cmd is available on non-cloud variants

simondeziel:
This is mainly to enable the cloud-init testing of desktop images for which we only build a single variant called `desktop`.

### DIFF
--- a/bin/test-image
+++ b/bin/test-image
@@ -84,7 +84,7 @@ if [ "${TYPE}" = "vm" ]; then
     INSTANCES="${TEST_IMAGE}"
 
     # Cloud-init testing.
-    if [ "${VARIANT}" = "cloud" ]; then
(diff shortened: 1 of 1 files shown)


MusicDin (approved):
Thanks

bin/test-image
```diff
@@ -84,7 +84,7 @@ if [ "${TYPE}" = "vm" ]; then
     INSTANCES="${TEST_IMAGE}"
 
     # Cloud-init testing.
-    if [ "${VARIANT}" = "cloud" ]; then
+    if [ "${VARIANT}" = "cloud" ] || lxc exec "${name}" -- sh -c "command -v cloud-init"; then
```
MusicDin, line 87: Missed that one: `${name}` -> `${TEST_IMAGE}`

Also, below is one more occurrence where cloud image is tested (for privileged containers)

PR #336, opened 2024-10-28
images/arch: Fix cloud-init enabled service

MusicDin:
Instead of enabling `cloud-init.service` enable `cloud-init-main.service`.

### DIFF
--- a/images/archlinux.yaml
+++ b/images/archlinux.yaml
@@ -824,7 +824,7 @@ actions:
     #!/bin/sh
     set -eux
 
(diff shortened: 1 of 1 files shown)


simondeziel (approved):

tomponline (commented):
What's the difference out of interest @MusicDin

MusicDin:
Service unit got renamed in new versions. This is a second distro where this surfaced, so we are expecting such an update for all images eventually.

PR #337, opened 2024-10-29
bin/test-image: use the right variable to test for cloud-init detection

simondeziel:


tomponline (approved):

MusicDin (approved):
Thanks. My comment yesterday included that cloud variants are also checked for containers, but I've forgot that we are only building desktop images for VMs.

bin/test-image
```diff
@@ -90,7 +90,13 @@ else
         INSTANCES="${INSTANCES} ${TEST_INSTANCE}"
 
         # FIXME: workaround for Linux 6.6.3 apparmor regression.
-        printf "2\ndenylist\nreject_force_umount\n[all]\nfsconfig errno 38\nfsopen errno 38\n" | lxc config set "${TEST_INSTANCE}" raw.seccomp -
+        lxc config set "${TEST_INSTANCE}" raw.seccomp "2
```
MusicDin, line 93: Could `2` be in a new line - for readability?

MusicDin (commented):
Should we also check cloud-init on reboot for desktop images?

I think extending the condition with the check for cloud-init should be enough:
```sh
# Check that cloud variants can be rebooted cleanly.
if [ "${VARIANT}" = "cloud" ]; then # <--
  ...
fi
```

MusicDin (approved):
Thanks 🚀

PR #341, opened 2024-11-04
tests/snapd: lxc benchmark

simondeziel:


tests/snapd
```diff
@@ -14,13 +14,6 @@ lxd waitready --timeout=300
 # Check that preseed via snap works
 [ "$(lxc config get core.https_address)" = "[::]:8443" ]
 
-# LXD setup
-lxc storage create default dir
```
tomponline, line 18: yeah this kinds of defeats the purpose of the test to setup LXD in advance. Good spot.

tests/snapd
```diff
@@ -30,15 +23,5 @@ su ${TEST_USER} -c "${LXC} info"
 echo "==> Deleting test user"
 userdel -r ${TEST_USER} 2>/dev/null || true
 
-echo "==> Deleting network device"
```
tomponline, line 33: can you keep this, but change it to clean up the resources that lxd-user should have created, that way if they dont exist we can detect the regression.

simondeziel: That's not my understanding of what the test does. Since the `${TEST_USER}` is made a member of the `lxd` group, it's not relying on `lxd-user`.

I'm happy to extend that to also test `lxd-user` but want to check with you first.

tomponline (approved):

tests/snapd
simondeziel: But since you've mentioned `lxd-user`, I want it to be covered as well, at the very least for code coverage reasons ;)
I'll fill up a Jira card to that effect.

PR #344, opened 2024-11-05
images/alt: Disable `update_hostname` cloud-init module

MusicDin:
Disable `update_hostname` module in cloud-init to prevent errors on reboot.

Passing build: https://github.com/MusicDin/lxd-ci/actions/runs/11682041046

### DIFF
--- a/images/alt.yaml
+++ b/images/alt.yaml
@@ -182,5 +182,17 @@ actions:
   variants:
   - cloud
 
+# XXX: Prevent cloud-init from attempting to modify the hostname. Otherwise, cloud-init fails with
+# an attribute error: "module 'cloudinit.util' has no attribute 'load_file'") in "update_hostname".
+- trigger: post-packages
+  action: |-
+    #!/bin/sh
+    set -eux
+
+    # Disable the update_hostname module in cloud-init.
+    sed -i '/update_hostname/d' /etc/cloud/cloud.cfg
+  variants:
+  - cloud
+
 mappings:
   architecture_map: altlinux


tomponline (approved):

images/alt.yaml
```diff
@@ -182,5 +182,17 @@ actions:
...
 
+# XXX: Prevent cloud-init from attempting to modify the hostname. Otherwise, cloud-init fails with
+# an attribute error: "module 'cloudinit.util' has no attribute 'load_file'") in "update_hostname".
+- trigger: post-packages
+  action: |-
+    #!/bin/sh
+    set -eux
+
+    # Disable the update_hostname module in cloud-init.
+    sed -i '/update_hostname/d' /etc/cloud/cloud.cfg
```
simondeziel, line 193: I wanted to find a way to drop a `/etc/cloud/cloud.cfg.d/` snippet to override that but AFAICT, that's not possible or at least not easily done while avoiding to redefine the `cloud_init_modules` list.

simondeziel (approved):

holmanb:
The change here probably makes sense to unblock image build, but it does leave cloud-init in an unexpected state. Do we know what version of cloud-init is installed in these images?

I've never seen anything like this before - I would assume from first glance that this is caused a broken downstream patch. Can someone please file a bug with ALT so that they know that this is happening?

MusicDin:
@holmanb Thanks for checking this one. The version of cloud-init is `24.2-alt1`: https://packages.altlinux.org/en/sisyphus/srpms/cloud-init/

holmanb:
> @holmanb Thanks for checking this one. The version of cloud-init is `24.2-alt1`: https://packages.altlinux.org/en/sisyphus/srpms/cloud-init/

I just did a little bit of digging, and it looks like this is definitely a downstream bug. ALT hasn't upstreamed their distro implementation. The source of the exception is from code that is in a file that doesn't exist in upstream cloud-init.

```python
Traceback (most recent call last):
  File "/usr/lib/python3/site-packages/cloudinit/config/cc_update_hostname.py", line 62, in handle
    cloud.distro.update_hostname(hostname, fqdn, prev_fn)
  File "/usr/lib/python3/site-packages/cloudinit/distros/__init__.py", line 546, in update_hostname
    prev_hostname = self._read_hostname(prev_hostname_fn)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/site-packages/cloudinit/distros/altlinux.py", line 154, in _read_hostname
    return util.load_file(filename).strip()
           ^^^^^^^^^^^^^^ 
```
 The fix for this is trivial, but I don't know where the repository is that implements `altlinux.py`. I just [filed a bug downstream](https://bugzilla.altlinux.org/52029).

simondeziel:
@holmanb thanks for the analysis and reporting it downstream!

holmanb:
@MusicDin @simondeziel fyi downstream just reported it was fixed:  https://bugzilla.altlinux.org/52029

PR #349, opened 2024-11-14
Cleanup unused images

MusicDin:
With @simondeziel we have checked the metrics for images, and this PR drops builds for unused images, especially focusing on distributions that are very *heavy* to build, such as Funtoo, and openEuler.

This removes builds for:
- Devuano (arm64 only)
- Gentoo (arm64 only)
- Funtoo (arm64 only)
- openEuler 22.03 (all)
- openEuler 24.03 (arm64 only)
- Oracle 8 (arm64 only)
- ALT p10 (all)

### DIFF
--- a/.github/workflows/image-alt.yml
+++ b/.github/workflows/image-alt.yml
@@ -25,7 +25,6 @@ jobs:
         release:
           - Sisyphus
           - p11
-          - p10
         variant:
           - default
           - cloud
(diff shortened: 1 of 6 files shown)


tomponline (approved):
Will this update the image website automatically?

MusicDin:
> Will this update the image website automatically?

The warning will show after 8 days (Last image build is older than 8 days), and it will disappear after 15 days (along with images). We do not have anything in place currently to explicitly tell the image has been deprecated and that it will no longer be updated.

tomponline:
We should post this on https://discourse.ubuntu.com/t/new-lxd-image-server-available-images-lxd-canonical-com/43824

PR #357, opened 2024-11-20
Make `tests/conversion` a bit faster

simondeziel:
Avoiding the 2 go snap installation should shave ~1m15s.

### DIFF
--- a/tests/conversion
+++ b/tests/conversion
@@ -20,8 +20,10 @@ install_lxd
 
 set -x
 
-# Install go from Snap.
-snap install go --classic
+if ! command -v go; then
+    # Install go from Snap.
+    snap install go --classic
+fi
 
 export PATH="${HOME}/go/bin:${PATH}"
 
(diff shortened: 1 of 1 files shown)


simondeziel:
> Avoiding the 2 go snap installation should shave ~1m15s.

The savings were actually more like 3-4 minutes but I guess that's mostly dependent on the snap store being slow or fast.

The CI failures are not related.

tests/conversion
```diff
@@ -255,7 +258,7 @@ IMAGE_PATH="${tmpdir}/backup/virtual-machine.img"
 # Test VM migration using conversion mode. If server does not support
 # conversion API extension, lxd-migrate must fallback to migration
 # mode and successfully transfer the VM disk.
-for driver in btrfs ceph dir lvm zfs; do
```
MusicDin, line 258: Why is `ceph` removed?

MusicDin (approved):
LGTM, just one question :) Thanks!

tests/conversion
simondeziel: Cause I didn't bother setting up microceph for testing locally and accidentally scooped that edit in... thanks for catching!

simondeziel: Oh incidentally, me accidentally dropping `ceph` from the tested storage driver probably explains why the tests were faster ... silly me!

tomponline (approved):

PR #358, opened 2024-11-20
images/almalinux: Fix almalinux mirror

MusicDin:
Fix mirror URL


simondeziel (approved):
Thanks for fixing this embarrassing and glaring error I introduced!

PR #360, opened 2024-11-20
images/openwrt: Install sudo on snapshot release

MusicDin:
Just follow up on https://github.com/canonical/lxd-ci/pull/359

### DIFF
--- a/images/openwrt.yaml
+++ b/images/openwrt.yaml
@@ -214,8 +214,6 @@ packages:
   - packages:
     - sudo
     action: install
-    releases:
-    - 23.05
 
 actions:
 - trigger: post-unpack
--- a/images/openwrt.yaml
+++ b/images/openwrt.yaml
@@ -224,6 +224,15 @@ actions:
   architectures:
   - amd64
 
+- trigger: post-unpack
+  action: |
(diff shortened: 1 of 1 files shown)


images/openwrt.yaml
```diff
@@ -214,8 +214,6 @@ packages:
   - packages:
     - sudo
     action: install
-    releases:
-    - 23.05
```
simondeziel, line 218: Looking at https://github.com/MusicDin/lxd-ci/actions/runs/11941866814, do I understand it right that the `manager: apk` bit is totally ignored and whatever is the package manager is used to install `sudo` as it's present in both `23.05` and `snapshot`?

simondeziel (approved):

images/openwrt.yaml
MusicDin: No, the used package manager in snapshot is `apk` in `23.05` is `opkg`. In the first case, the package manager from config is used, in the latter from the lxd-imagebuilder flag.

MusicDin: It is not ignored. We need to tell lxd-imagebuilder which manager to use, as it wraps install/update/remove commands for different managers.

PR #362, opened 2024-11-26
CI tweaks

simondeziel:


simondeziel:
@tomponline https://github.com/canonical/lxd-ci/actions/runs/12035105813/job/33553318964?pr=362#step:11:1115 looks suspicious as the stop hook is just not returning inside the allocated 10m.

tomponline:
> @tomponline https://github.com/canonical/lxd-ci/actions/runs/12035105813/job/33553318964?pr=362#step:11:1115 looks suspicious as the stop hook is just not returning inside the allocated 10m.

Can we see the lxd debug logs (they dont seem to be in the lxd logs section)

simondeziel:
It seems to have been a fluke as now it's stopping in ~1m.

> Can we see the lxd debug logs (they dont seem to be in the lxd logs section)

That is correct, debug logging is not enabled. Should I add the needed `snap set lxd` bit in `bin/helpers`' `install_lxd()` to unconditionally enable debug logging?

tomponline:
> It seems to have been a fluke as now it's stopping in ~1m.
> 
> > Can we see the lxd debug logs (they dont seem to be in the lxd logs section)
> 
> That is correct, debug logging is not enabled. Should I add the needed `snap set lxd` bit in `bin/helpers`' `install_lxd()` to unconditionally enable debug logging?

Yeah we could try that and see how it goes

tomponline (approved):
Ta!

PR #366, opened 2024-12-05
tests/storage-volumes-vm: Root volume disk device attachments

MggMuggins:
This should have a check for all corner cases around VM root volume attachments (coverage for https://github.com/canonical/lxd/pull/14532):
- security.protection.start allows one other VM to attach the machine's root disk, and can only be removed if the disk is not attached
- security.shared allows unchecked attachments of root disks
- VM attachments are correctly reported in used_by
- hotplug of VM root attachments works (as this is the method reccomended by the docs to avoid UUID/LABEL conflicts)

I ran into an interesting behavior of udev in Noble VMs while testing that slowed this down a bit. In order for udev to create a symlink for `SCSI_IDENT_SERIAL` (`/dev/disk/by-id/scsi-SQEMU_QEMU_HARDDISK_lxd_virtual--machine-vm5`), I had to hotplug the device in an `ubuntu:24.04` VM. Starting the VM from cold didn't create the link, and `ubuntu-minimal:24.04` VMs never created it at all.

Without hotplug or in an `ubuntu-minimal:24.04`:
```
# udevadm info /dev/sdb
...
S: disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_virtual--machine
S: disk/by-path/pci-0000:02:00.0-scsi-0:0:1:1
S: disk/by-diskseq/10
E: DEVPATH=/devices/pci0000:00/0000:00:01.1/0000:02:00.0/virtio6/host0/target0:0:1/0:0:1:1/block/sdb
E: DEVNAME=/dev/sdb
...
E: ID_SERIAL=0QEMU_QEMU_HARDDISK_lxd_virtual--machine
E: ID_SERIAL_SHORT=lxd_virtual--machine
E: ID_SCSI_SERIAL=lxd_virtual--machine-v2
E: ID_BUS=scsi
...
E: DEVLINKS=/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_virtual--machine /dev/disk/by-path/pci-0000:02:00.0-scsi-0:0:1:1 /dev/disk/by-diskseq/10
...
```
With hotplug:
```
# udevadm info /dev/sdb
...
S: disk/by-id/scsi-SQEMU_QEMU_HARDDISK_lxd_virtual--machine-vm5
S: disk/by-path/pci-0000:02:00.0-scsi-0:0:1:1
S: disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_virtual--machine
S: disk/by-diskseq/15
...
E: DEVPATH=/devices/pci0000:00/0000:00:01.1/0000:02:00.0/virtio6/host0/target0:0:1/0:0:1:1/block/sdb
E: DEVNAME=/dev/sdb
...
E: SCSI_IDENT_SERIAL=lxd_virtual--machine-vm5
E: SCSI_IDENT_LUN_VENDOR=lxd_virtual--machine
...
E: ID_SERIAL=0QEMU_QEMU_HARDDISK_lxd_virtual--machine
E: ID_SERIAL_SHORT=lxd_virtual--machine
E: ID_SCSI_SERIAL=lxd_virtual--machine-vm5
...
E: DEVLINKS=/dev/disk/by-id/scsi-SQEMU_QEMU_HARDDISK_lxd_virtual--machine-vm5 /dev/disk/by-path/pci-0000:02:00.0-scsi-0:0:1:1 /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_virtual--machin>
...
```

I worked around it by using a shorter LXD device name so that the symlink name wouldn't be truncated; not sure if this is worth chasing further.

### DIFF
--- a/tests/storage-volumes-vm
+++ b/tests/storage-volumes-vm
@@ -191,6 +191,86 @@ do
 	lxc storage volume detach "${poolName}" vol3 v1
 	lxc storage volume detach "${poolName}" vol6 v1 || true  # optional ISO
 
+	# attach VM root volumes
+	if hasNeededAPIExtension instance_root_volume_attachment; then
+		empty_vm_size=8KiB
+		if [ "${poolDriver}" = "powerflex" ]; then
+			empty_vm_size=8GiB
+		fi
+
+		lxc init --empty --vm v2 --storage "${poolName}" --device root,size="${empty_vm_size}"
+		lxc init --empty --vm v3 --storage "${poolName}" --device root,size="${empty_vm_size}"
+
+		# Requires either security.shared or security.protection.start
+		! lxc storage volume attach "${poolName}" virtual-machine/v2 v1 || false
+
+		lxc config set v2 security.protection.start=true
+
+		# security.protection.start on a VM allows exactly one other attachment
+		lxc storage volume attach "${poolName}" virtual-machine/v2 v1
+		! lxc storage volume attach "${poolName}" virtual-machine/v2 v3 || false
+
+		# Deleting the instance will fail while it's root volume is in use
+		! lxc delete v2 || false
+
+		# Make sure used_by is calculated correctly
+		lxc storage volume show "${poolName}" virtual-machine/v2 | grep -qF '/1.0/instances/v1'
+
+		# Can't unset security.protection.start when v2's root volume is attached to vm1
+		! lxc config unset v2 security.protection.start || false
+
+		lxc storage volume detach "${poolName}" virtual-machine/v2 v1
+
+		# Unset security.protection.start works when not attached
+		lxc config unset v2 security.protection.start
+
+		lxc config set v2 security.protection.start=true
+		lxc storage volume attach "${poolName}" virtual-machine/v2 v1
+
+		lxc storage volume set "${poolName}" virtual-machine/v2 security.shared=true
+
+		# Unset security.shared works when security.protection.start=true
+		lxc storage volume unset "${poolName}" virtual-machine/v2 security.shared
+		lxc storage volume set "${poolName}" virtual-machine/v2 security.shared=true
+
+		# security.shared allows many attachments
+		lxc storage volume attach "${poolName}" virtual-machine/v2 v3
+
+		lxc storage volume show "${poolName}" virtual-machine/v2 | grep -qF '/1.0/instances/v1'
+		lxc storage volume show "${poolName}" virtual-machine/v2 | grep -qF '/1.0/instances/v3'
+
+		lxc config unset v2 security.protection.start
+
+		# Detach so that we can double-check hotplug
+		lxc storage volume detach "${poolName}" virtual-machine/v2 v1
+
+		# Make sure that the devices actually show up and can be mounted
+		lxc start v1
+		waitInstanceReady v1
+
+		# Specify a different device name here; udev appears to be truncating the
+		# default name when it creates the /dev/disk/by-id/scsi* symlinks, so use
+		# a shorter name to prevent truncation.
+		lxc storage volume attach "${poolName}" virtual-machine/v2 v1 v2-root
+		lxc exec v1 -- test -L /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_v2--root
+		lxc stop --force v1
+
+		# Can't unset security.shared when v1's root volume is attached elsewhere
+		! lxc storage volume unset "${poolName}" virtual-machine/v2 security.shared || false
+
+		# Instances cannot be moved when their root disk is attached to another instance
+		! lxc move v2 v4 || false
+
+		lxc storage volume detach "${poolName}" virtual-machine/v2 v1
+		lxc storage volume detach "${poolName}" virtual-machine/v2 v3
+
+		lxc storage volume unset "${poolName}" virtual-machine/v2 security.shared
+
+		lxc delete v2 v3
+	else
+		echo "==> Skipping instance root attachment tests, not supported"
+	fi
+
 	echo "==> Deleting VM"
 	lxc delete v1
 


tests/storage-volumes-vm
```diff
@@ -191,6 +191,74 @@ do
...
+
+		# Can't unset security.shared when vm1's root volume is attached elsewhere
+		! lxc storage volume unset "${poolName}" virtual-machine/v2 security.shared || false
+
+		lxc storage volume detach "${poolName}" virtual-machine/v2 v1
+		lxc storage volume detach "${poolName}" virtual-machine/v2 v3
+
+		lxc storage volume unset "${poolName}" virtual-machine/v2 security.shared
+
+		lxc delete --force v2 v3
```
simondeziel, line 257: Force shouldn't be needed as they are not running. To be sure force isn't required, could we lose it?

tests/storage-volumes-vm
```diff
@@ -191,6 +191,74 @@ do
...
+		! lxc storage volume attach "${poolName}" virtual-machine/v2 v1 || false
+
+		lxc config set v2 security.protection.start=true
+
+		# security.protection.start on a VM allows exactly one other attachment
+		lxc storage volume attach "${poolName}" virtual-machine/v2 v1
+		! lxc storage volume attach "${poolName}" virtual-machine/v2 v3 || false
+
+		# Deleting the instance will fail while it's root volume is in use
+		! lxc delete v2 || false
```
simondeziel, line 209: I am now wondering if you can remove `v2`'s root volume and permanently attach it to `v1`? I think that's something we want for scenario where we "rescue" files from an unbootable VM.

The test is fine but that question that popped :)

simondeziel (commented):
LGTM, I made some comments but they are mostly nit picks.

I haven't seen it run in GHA due to the extension not being in `latest/edge` just yet. I'm assuming it all works locally but wanted to check how much longer it made the test :)

tomponline:
> I ran into an interesting behavior of udev in Noble VMs while testing that slowed this down a bit. In order for udev to create a symlink for `SCSI_IDENT_SERIAL` (`/dev/disk/by-id/scsi-SQEMU_QEMU_HARDDISK_lxd_virtual--machine-vm5`), I had to hotplug the device in an `ubuntu:24.04` VM. Starting the VM from cold didn't create the link, and `ubuntu-minimal:24.04` VMs never created it at all.

@hamistao any experience with this?

hamistao:
Not particularly, but it is not the first time we have seen udev or other packages mangling with disk devices that LXD manages (https://github.com/canonical/lxd/issues/13701#issuecomment-2231983935) and since it only happens on a specific distro version and not on minimal it is almost certainly a problem of the same nature and probably isn't worth digging too deep into.

On `Without hotplug or in an ubuntu-minimal:24.04` we have `S: disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_virtual--machine` and `E: DEVLINKS=/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_virtual--machine ...` so LXD seems to be doing its job just fine.

tests/storage-volumes-vm
MggMuggins: With this feature, the only meaningful distinction remaining between `virtual-machine`/`container` volumes and `custom` volumes is their lifecycle; that is, tied to instances vs independent. In the absence of that constraint (and any concerns around back-compat :sweat_smile: ), I'd advocate to remove the idea of storage volume types all together and just have "storage volumes".

If keeping the root's contents around without the instance is a capability that we want, it might be worth looking at whether creating a custom volume as a copy of an instance/snapshot is possible.

MggMuggins:
Thanks @tomponline @hamistao, I agree that this isn't worth chasing.

@simondeziel They do pass locally :wink: . It looks like it goes from 5 `lxc start`s to 6 `lxc start`s. So ~120% of the previous runtime for this test.

tests/storage-volumes-vm
```diff
@@ -191,6 +191,74 @@ do
 	lxc storage volume detach "${poolName}" vol3 v1
 	lxc storage volume detach "${poolName}" vol6 v1 || true  # optional ISO
 
+	# attach VM root volumes
+	if hasNeededAPIExtension instance_root_volume_attachment; then
+		lxc init --empty --vm v2 --storage "${poolName}" --device root,size=8KiB
+		lxc init --empty --vm v3 --storage "${poolName}" --device root,size=8KiB
```
simondeziel, line 197: It's good to test with such small volumes. I'm pretty sure LVM will round them up but that should happen all transparently, hopefully!

Dunno how it would fare with PowerFlex's minimum of 8GiB though. I guess we'll see when we run it there :)

simondeziel (dismissed):

tests/storage-volumes-vm
tomponline: powerflex doesnt round up to 8gib afaik @roosterfish ?

MggMuggins:
Added a check to make sure that `security.shared` can be unset when `security.protection.start` is set.

MggMuggins:
Added a check for instance rename.

tests/storage-volumes-vm
MggMuggins: Pushed a special case for powerflex.

tomponline:
@MggMuggins seems to be failing:

> + lxc storage volume attach vmpool14779 virtual-machine/v2 v1 v2-root
Error: Failed to start device "v2-root": Failed to call monitor hook for block device: Failed adding block device for disk device "v2-root": Failed adding block device: error reading header from custom_test_v2.block: No such file or directory

MggMuggins:
The qemu driver infers the volume type based on the source path ([here](https://github.com/canonical/lxd/blob/main/lxd/instance/drivers/driver_qemu.go#L3989)). It seems to me that the "RBDFormat" (see [here](https://github.com/canonical/lxd/blob/main/lxd/device/device_utils_disk.go#L55)) should really be sending over the volume name as it appears to Ceph, rather than requiring the driver to figure out project and volume type info.

I'm working on building a reproducer and should have a fix up by EOD Wed.

MggMuggins:
https://github.com/canonical/lxd/pull/14807 fixes this.

tomponline (approved):
Thanks!

tomponline:
@MggMuggins this will need to be updated now too since https://github.com/canonical/lxd/pull/14886 right?

MggMuggins:
Sorry, just saw this. As we discussed last week, https://github.com/canonical/lxd/pull/14886 only changes the device config keys used by the server; the cli interface didn't change, so this is alright without updates.

PR #330, opened 2024-10-25
tests/network-ovn: Add MicroOVN as a backend for network-ovn test suite

gabrielmougard:
related to https://github.com/canonical/lxd/pull/14327

PR #368, opened 2024-12-11
Test against both MicroOVN and OVN debs

simondeziel:
Supersedes #330

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -97,7 +97,8 @@ jobs:
           - lxd-user
           - interception
           - network-bridge-firewall
-          - network-ovn
+          - network-ovn ovn:deb
+          - network-ovn ovn:latest/edge
           - network-routed
           - pylxd
           - snapd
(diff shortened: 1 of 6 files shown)


simondeziel:
@tomponline as we know `vm-migration` are not working (reliably) ATM but it's got nothing to do with this PR that barely touches it to avoid wasting too much time on a hung migration.

`tl;dr`: ready for review/merge when you are

.github/workflows/tests.yml
```diff
@@ -203,6 +208,10 @@ jobs:
             track: "5.0/candidate"
           - test: efi-vars-editor-vm
             track: "5.0/edge"
+          - test: network-ovn ovn:latest/edge
```
tomponline, line 211: these last 2 variants stood out to me as different to the rest as they dont include a deb variant?

simondeziel: The review UI doesn't make it easy to spot with our overly long exclusion matrix but yeah, that's an exclusion. So with 5.0, we test only the `ovn:deb` variant hence the exclusion of the `ovn:latest/edge` one/

tomponline (approved):

PR #372, opened 2024-12-19
Turn microceph setup into a reusable helper

simondeziel:
This is the first step toward not relying on GH actions to setup MicroCeph's snap. It replaces the ad-hoc manual setup by a set of 2 helper functions.

For now, it's only used to replace the `ceph` VM from the `tests/vm-migration` by a `microceph` snap deployed in the GHA runner VM. For now it doesn't use the ephemeral disk but a loop-backed file.

### DIFF
--- a/bin/helpers
+++ b/bin/helpers
@@ -141,6 +141,62 @@ install_deps() (
     fi
 )
 
+# install_microceph: install MicroCeph snap.
+install_microceph() (
+    if snap list microceph 2>/dev/null; then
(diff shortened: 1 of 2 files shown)


simondeziel:
@tomponline this only shaves 20-30 seconds of runtime only but removing one of the 3 VMs good on its own. This opens the door to bumping the RAM of `member1`/`member2`.

tomponline:
Hopefully we can remove the other 2 VMs as well and use containers.

PR #385, opened 2025-01-13
tests/cgroup: avoid prompts when installing iperf3

simondeziel:
Fixes https://github.com/canonical/lxd/issues/14772


tomponline (approved):

tomponline:
@simondeziel @mihalicyn We are seeing this happening more often in recent weeks, i took a quick look and it appears an inability to get rstatus from the pipe. Running the command again succeeds. Ive experienced this myself locally intermittently too.

PR #386, opened 2025-01-14
Setup swapfile in `tests/cgroup` if needed

simondeziel:


simondeziel:
@tomponline tested locally on a freshly create LXD VM lacking swap.

PR #388, opened 2025-01-22
actions/image-upload: Fix env field location

MusicDin:
The `env` key cannot be located under `runs` in composite actions.


tomponline (approved):

simondeziel:
Sorry for this regression and thanks for fixing it!

PR #394, opened 2025-01-23
github: change how qemu-external-vm test is added to the test matrix

simondeziel:
The problem with the `include` set is that if one asks to run all tests for say `5.21/edge` on `22.04`, the `include` set will also be added to the mix which is not desired.


tomponline (approved):
Thanks!

PR #398, opened 2025-01-28
bin/helpers: Update snapd before installing lxd

tomponline:


gabrielmougard (approved):
LGTM

simondeziel:
@tomponline your previous attempt that looked like like this:

```
 # install_lxd: install LXD from a specific channel or `latest/edge` if none is provided.
 # Optional argument: boolean which indicates whether to start the daemon. Default is true.
 install_lxd() (
     local start_daemon="${1:-true}"
 
+    # Ensure latest snapd in case LXD requires a newer one than is installed.
+    apt-get update --yes || true
+    snap refresh snapd --cohort="+" || true
+
     # Wait for snapd seeding
     waitSnapdSeed

```

Looked promising to me but the `apt-get update` part was not necessary as you wanted to refresh a snap, not a dep. The `snap refresh snapd` should have been after the `waitSnapdSeed` I think.

This should have work, I think. I'll send a PR to that effect.

tomponline:
> The `snap refresh snapd` should have been after the `waitSnapdSeed` I think.

I figured to do it before otherwise we probably need to wait for it to seed again right?

simondeziel:
> > The `snap refresh snapd` should have been after the `waitSnapdSeed` I think.
> 
> I figured to do it before otherwise we probably need to wait for it to seed again right?

Dunno about having to wait for the seed again, seems reasonable to me but if the initial snapd is not seeded, it will just refuse any refresh operation.

tomponline:
> > > The `snap refresh snapd` should have been after the `waitSnapdSeed` I think.
> > 
> > 
> > I figured to do it before otherwise we probably need to wait for it to seed again right?
> 
> Dunno about having to wait for the seed again, seems reasonable to me but if the initial snapd is not seeded, it will just refuse any refresh operation.

I was seeing errors about snapd not being installed when trying to refresh it, but this change of making it happen after seeding seems to have avoided that. Great!

simondeziel:
> > > > The `snap refresh snapd` should have been after the `waitSnapdSeed` I think.
> > > 
> > > 
> > > I figured to do it before otherwise we probably need to wait for it to seed again right?
> > 
> > 
> > Dunno about having to wait for the seed again, seems reasonable to me but if the initial snapd is not seeded, it will just refuse any refresh operation.
> 
> I was seeing errors about snapd not being installed when trying to refresh it, but this change of making it happen after seeding seems to have avoided that. Great!

Hmm, this is an intentional side effect so I've been explicit with the `install`/`refresh` behaviour instead to avoid relying on that fact.

The dual mechanism of shipping `snapd` through deb and snap brings subtle complexities :/

PR #402, opened 2025-01-31
tests/conversion: workaround systemd-resolved not picking up the edit to /etc/hosts in time

simondeziel:


MusicDin (approved):

MusicDin:
Just for my understanding, why is the potential error from `resolvectl flush-caches || true` ignored? Can this command fail if already flushed?

simondeziel:
> Just for my understanding, why is the potential error from `resolvectl flush-caches || true` ignored? Can this command fail if already flushed?

It indeed serves no purpose so I dropped it. Thanks!

tomponline (approved):

PR #403, opened 2025-01-31
tests: Test mount options for hotplugged dir share device

hamistao:
Tests for https://github.com/canonical/lxd/pull/14898. I am assuming it will be backported into 5.21. If not, let me know and I shall update the tests accordingly.

### DIFF
--- a/tests/storage-disks-vm
+++ b/tests/storage-disks-vm
@@ -141,7 +141,13 @@ if echo "${LXD_SNAP_CHANNEL}" | grep -E '^([45]\.0)/'; then
   ! lxc config device add v1 dir1 disk source="${testRoot}/allowed1" path="/mnt/bar" || false
 else
   # Hotplugging directories is allowed from LXD 5.21 onwards
-  lxc config device add v1 dir1 disk source="${testRoot}/allowed1" path="/mnt/bar"
+  lxc config device add v1 dir1 disk source="${testRoot}/allowed1" path="/mnt/bar" readonly=true
+  sleep 1
+  lxc exec v1 -- mount | grep -w "lxd_dir1 on /mnt/bar type virtiofs" | grep -w ro
+  lxc config device set v1 dir1 readonly=false
+  sleep 1
+  lxc exec v1 -- mount | grep -w "lxd_dir1 on /mnt/bar type virtiofs" | grep -w rw
+
   lxc config device remove v1 dir1
 fi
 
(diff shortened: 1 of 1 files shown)


tests/storage-disks-vm
```diff
@@ -141,7 +141,15 @@ if echo "${LXD_SNAP_CHANNEL}" | grep -E '^([45]\.0)/'; then
   ! lxc config device add v1 dir1 disk source="${testRoot}/allowed1" path="/mnt/bar" || false
 else
   # Hotplugging directories is allowed from LXD 5.21 onwards
-  lxc config device add v1 dir1 disk source="${testRoot}/allowed1" path="/mnt/bar"
+  lxc config device add v1 dir1 disk source="${testRoot}/allowed1" path="/mnt/bar" readonly=true
+  sleep 1
+  DISK_DEV="$(basename "$(lxc exec v1 -- readlink -f /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_dir1)")"
+  lxc exec v1 -- sh -c "for i in \$(seq 10); do test -e /sys/block/${DISK_DEV}/ro && break; echo \"Waiting for sys file to appear (\${i}s)\"; sleep 1; done"
```
simondeziel, line 147: This never seems to break in tests: https://github.com/canonical/lxd-ci/actions/runs/13078926811/job/36497780387#step:12:662

tests/storage-disks-vm
```diff
@@ -141,7 +141,15 @@ if echo "${LXD_SNAP_CHANNEL}" | grep -E '^([45]\.0)/'; then
...
+  lxc exec v1 -- mount | grep dir1 | grep ro
```
simondeziel, line 148: This seems to match on leftovers from previous tests:

```
+ lxc exec v1 -- mount
+ grep dir1
+ grep rw
lxd_dir1ro on /srv/ro type virtiofs (rw,relatime)
lxd_dir1rw on /srv/rw type virtiofs (rw,relatime)
lxd_dir1ro on /srv/ro9p type 9p (rw,relatime,access=client,trans=virtio)
```
Note that those are not mounted to `/mnt/bar` as they should be, isn't it?

simondeziel (commented):
If we expect a block device, I'd suggest to check the content of `/sys/block/${DISK_DEV}/ro` for `1` when in read-only and `0` when in read-write mode` instead (or in addition) of checking the output of `mount`.

tests/storage-disks-vm
hamistao: Perhaps once I add the test for `/sys/block/${DISK_DEV}/ro` we will have a better clue of why that is

tests/storage-disks-vm
hamistao: I just added a grep too permissive here, thanks for catching that. The new entry that would be mounted on /mnt/bar should not be there yet as the PR with the necessary changes was not merged yet.

tests/storage-disks-vm
hamistao: In fact we are not expecting a block device here, since this is a directory share device and virtiofs apparentely does not use traditional block devices.

tomponline (approved):

PR #405, opened 2025-02-03
tests/conversion: restart systemd-resolved after altering /etc/hosts

simondeziel:


simondeziel:
This kludge seems a necessary evil as the loop ran 17 times (with `sleep 0.1`) before `getent hosts` actually returned the proper DNS address.

tomponline (commented):
Did you try restarting the systemd-resolved service?

simondeziel:
@tomponline all green now, thanks for the restart suggestion, much cleaner.

tomponline (approved):

PR #408, opened 2025-02-04
tests/network: Remove invalid `mtu` device option for physical NIC `agent.nic_config` test

kadinsayani:


tests/network
```diff
@@ -155,7 +155,7 @@ echo "=> Performing VM physical NIC agent.nic_config tests"
 lxc exec v1-physical -- "sync"
 lxc stop v1-physical -f
 lxc config set v1-physical agent.nic_config=true
-lxc config device set v1-physical eth0 nictype=physical parent="${parentNIC}v1" mtu=1400 name=eth0
+lxc config device set v1-physical eth0 nictype=physical parent="${parentNIC}v1" name=eth0
```
tomponline, line 158: So does the lxd-agent not apply the mtu inside the VM when agent.nic_config=true?

simondeziel: Indeed, the `mtu` setting is currently being rejected as invalid.

tomponline: im getting a bit confused by your recent changes now, as we see its checking on this line https://github.com/canonical/lxd-ci/pull/408/files#diff-6b21ed85e37851843925564db3aef69588331fb257b2091c236e217ff428a4d1R169 suggesting it does support it (or at least did at one time recently).

kadinsayani: It does, but we can't pass `mtu` as a device option for VMs because of this check: https://github.com/canonical/lxd/blob/main/lxd/device/nic_physical.go#L34-L51.

tomponline: the tests were passing though recently (run it recently on 5.21/candidate for example). Can you double check locally?

tomponline: Oh i see you added this check recently https://github.com/canonical/lxd-ci/pull/404/files

kadinsayani: I don't think https://github.com/canonical/lxd/pull/14917/commits/0d412b8f077df6b84f477c2daf63a5bcfc81548a is a regression since those device options can't be set for physical NICs on VMs.

kadinsayani: The tests search for SRIOV enabled NICs:

https://github.com/kadinsayani/lxd-ci/blob/447d40974feae8de3e023d977894d1e13f2256d2/tests/network#L22-L29

simondeziel: I tried providing a non-SRIOV physical NIC to `v1-physical` and while booting the VM makes the NIC disappear from my host, it's nowhere to be seen in the VM:

```
$ ip link | grep enp2s0f0
22: enp2s0f0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc fq_codel state DOWN mode DEFAULT group default qlen 1000

$ lxc config show v1-physical | sed -n '/^devices:/,/^ephemeral/ p'
devices:
  eth0:
    name: enp5s0
    network: lxdbr0
    type: nic
  eth1:
    name: eth1
    nictype: physical
    parent: enp2s0f0
    type: nic
ephemeral: false
$ lxc start v1-physical
$ ip link | grep enp2s0f0
```

Then in the VM, we only see the first NIC, not the physical one:

```
$ lxc shell v1-physical 
root@v1-physical:~# ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: enp5s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group default qlen 1000
    link/ether 00:16:3e:fe:61:5a brd ff:ff:ff:ff:ff:ff
root@v1-physical:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute 
       valid_lft forever preferred_lft forever
2: enp5s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:16:3e:fe:61:5a brd ff:ff:ff:ff:ff:ff
    inet 172.24.26.49/24 metric 100 brd 172.24.26.255 scope global dynamic enp5s0
       valid_lft 3483sec preferred_lft 3483sec
    inet6 2001:470:b1c3:7946:54bd:2e11:44ff:3998/128 scope global dynamic noprefixroute 
       valid_lft 3486sec preferred_lft 3486sec
    inet6 fe80::216:3eff:fefe:615a/64 scope link 
       valid_lft forever preferred_lft forever
```

tomponline: @simondeziel do you need to install the drivers for the NIC in the guest?

kadinsayani:
Thanks @simondeziel for testing and collaborating on this one :)

tests/network
```diff
@@ -150,24 +150,23 @@ networkTests
...
-lxc config device set v1-physical eth0 nictype=physical parent="${parentNIC}v1" mtu=1400 name=eth0
+lxc config device add v1-physical eth1 nictype=physical parent="${parentNIC}v2" name=eth1
 lxc start v1-physical
 
 # Wait for lxd-agent to rename the interface.
 waitInstanceBooted v1-physical
 
 # Get the MAC address of the host NIC.
-parentNICMAC=$(cat /sys/class/net/"${parentNIC}v1"/address)
+parentNICMAC=$(cat /sys/class/net/"${parentNIC}v2"/address)
```
simondeziel, line 165: This needs to be read before `v1-physical` is started as when it starts, the device and it's associated `/sys` files are gone from the host.

tests/network
```diff
@@ -150,24 +150,23 @@ networkTests
...
 
-# Interface "eth0" should exist in the VM with the correct MTU and MAC address.
-lxc exec v1-physical -- test -d /sys/class/net/eth0
-lxc exec v1-physical -- grep -Fx 1400 /sys/class/net/eth0/mtu
-lxc exec v1-physical -- cat /sys/class/net/eth0/address | grep -Fx "${parentNICMAC}"
+# Interface "eth1" should exist in the VM with the correct name and MAC address.
+lxc exec v1-physical -- test -d /sys/class/net/eth1
+lxc exec v1-physical -- cat /sys/class/net/eth1/address | grep -Fx "${parentNICMAC}"
```
simondeziel, line 169:
```suggestion
[ "$(lxc exec v1-physical -- cat /sys/class/net/eth1/address)" = "${parentNICMAC}" ]
```

tests/network
simondeziel: I don't understand how it worked before:

```
$ cat /sys/class/net/enp2s0f0/address 
91:2a:c2:cc:a1:a2
$ lxc start v1-physical 
$ cat /sys/class/net/enp2s0f0/address
cat: /sys/class/net/enp2s0f0/address: No such file or directory
$ lxc delete -f v1-physical 
$ cat /sys/class/net/enp2s0f0/address
91:2a:c2:cc:a1:a2
```

tests/network
```diff
@@ -150,24 +150,27 @@ networkTests
 lxc exec v1-macvlan -- apt-get install --no-install-recommends --yes ethtool
 lxc exec v1-macvlan -- ethtool -l enp5s0 | grep -c '^Combined:\s\+3$'  | grep -Fx 2
 
-# Check VM "agent.nic_config" works by reconfiguring eth0 to use parent and mtu settings.
+# Check VM "agent.nic_config" works by adding "eth1" set to use parent name and MAC address.
```
simondeziel, line 153:
```suggestion
# Check VM "agent.nic_config" works by adding "eth1" to use parent name and MAC address.
```

tests/network
```diff
@@ -150,24 +150,27 @@ networkTests
...
 echo "=> Performing VM physical NIC agent.nic_config tests"
 lxc exec v1-physical -- "sync"
 lxc stop v1-physical -f
+
+# Get the MAC address of the host NIC before starting instance.
+parentNICMAC=$(cat /sys/class/net/"${parentNIC}v3"/address)
+
 lxc config set v1-physical agent.nic_config=true
-lxc config device set v1-physical eth0 nictype=physical parent="${parentNIC}v1" mtu=1400 name=eth0
+lxc config device set v1-physical eth1 nictype=physical parent="${parentNIC}v3" name=eth1
```
simondeziel, line 162:
```suggestion
lxc config device add v1-physical eth1 nictype=physical parent="${parentNIC}v3" name=eth1
```

tests/network
kadinsayani: Maybe `configured` instead of `set`?

tests/network
kadinsayani: Missed this while slicing up commits, thanks.

tests/network
kadinsayani, line 153:
```suggestion
# Check VM "agent.nic_config" works by adding "eth1" configured to use parent name and MAC address by `agent.nic_config`.
```

simondeziel: Ah, I'm fine with either. I initially thought it was to mirror the `lxc config device set` that slip by accident.

simondeziel (approved):
Thanks!

tomponline (approved):
Ta

PR #409, opened 2025-02-06
tests/storage-volumes-vm: Test VM disk device unmount

MggMuggins:
Tests for https://github.com/canonical/lxd/pull/14929

I have verified that these fail before applying those changes.

### DIFF
--- a/tests/storage-volumes-vm
+++ b/tests/storage-volumes-vm
@@ -254,6 +254,11 @@ do
 		lxc storage volume attach "${poolName}" virtual-machine/v2 v1 v2-root
 		sleep 3
 		lxc exec v1 -- test -L /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_v2--root
+
+		# `lxc stop [-f]` won't fail if unmounting the device fails
+		# `detach` will fail if the unmount fails, so just detach/reattach
+		lxc storage volume detach "${poolName}" virtual-machine/v2 v1
+		lxc storage volume attach "${poolName}" virtual-machine/v2 v1 v2-root
 		lxc stop --force v1
 
 		# Can't unset security.shared when v1's root volume is attached elsewhere


tests/storage-volumes-vm
```diff
@@ -254,6 +254,10 @@ do
 		lxc storage volume attach "${poolName}" virtual-machine/v2 v1 v2-root
 		sleep 3
 		lxc exec v1 -- test -L /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_v2--root
+
+		# Test that unmount succeeds; failed unmount will not cause VM stop to fail.
+		lxc storage volume detach "${poolName}" virtual-machine/v2 v1
```
simondeziel, line 259: CI is unhappy about this.

simondeziel (dismissed):

tests/storage-volumes-vm
MggMuggins: Yep, depends on https://github.com/canonical/lxd/pull/14929. Since that's merged we can rerun those for latest/edge once the build finishes.

tests/storage-volumes-vm
```diff
@@ -254,6 +254,10 @@ do
 		lxc storage volume attach "${poolName}" virtual-machine/v2 v1 v2-root
 		sleep 3
 		lxc exec v1 -- test -L /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_lxd_v2--root
+
+		# Test that unmount succeeds; failed unmount will not cause VM stop to fail.
```
simondeziel, line 258: If I followed correctly, this detach and reattach the same disk (`v2`'s root) to `v1`. If so, I don't understand the comment about a failed umount not impacting the VM stop.

Is it that `detach` will succeed even if `umount` fails from the instance? Or is it that a normal (non-force) `lxc stop v1` would work?

MggMuggins: `lxc stop` (with or without `-f`) won't fail if one of it's devices fails to stop.

Other way around: `detach` will fail if the `unmount` fails.

I pushed a clarification to the comment (hopefully).

simondeziel (approved):

tomponline:
@MggMuggins i re-ran this today with the latest/edge build apparently succeeding, so looks like the tests are still failing.

MggMuggins:
Documenting it here:
```
lxc config device remove <instance> <device>
Error: Failed to stop device "v1-snap0": Failed to deactivate LVM logical volume "/dev/default/virtual-machines_v1-snap0.block": Failed to run: lvchange --activate n --ignoreactivationskip /dev/default/virtual-machines_v1-snap0.block: exit status 5 (Logical volume default/virtual-machines_v1-snap0.block in use.)
```

After the command fails, the block device is completely gone from the VM, although the lvm symlink & dm block device are (ofc) still there.

LXD runs `lvchange` 20 times before finally giving up (see [this commit](https://github.com/canonical/lxd/commit/e72f7ff30ef450f7a00f53d689873fbaa0d67be3)). I upped this to 50 attempts but it still fails.

The failure prevents the disk device being removed from the DB, so shutting down the VM attempts to `lvchange` it again. I suspect this happens after the qemu process has ended. Shutdown produces no errors. 

Suspect that qemu/lxd is holding on to an FD as the block device isn't mounted anywhere.

More Monday.

MggMuggins:
This bug also affects lvm custom storage volumes, and I've just reproduced it in `6/stable` (`6.2-bde4d03`) as well.

tomponline:
@MggMuggins does it need fixing in 5.21/candidate too?

PR #412, opened 2025-02-11
tests: Introduce cloud-init tests

hamistao:
This includes the tests for `cloud-init.ssh-keys`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,6 +85,7 @@ jobs:
         track: ${{ fromJSON(inputs.snap-tracks || '["latest/edge", "5.21/edge", "5.0/edge", "4.0/edge"]') }}
         test:
           - cgroup
+          - cloud-init
           - cluster
           - container
           - container-copy
--- a/tests/cloud-init
+++ b/tests/cloud-init
@@ -1,0 +1,60 @@
+#!/bin/bash
+set -eux
+
+# Install LXD
+install_lxd
+
+if ! hasNeededAPIExtension "cloud_init"; then
+  echo "Skipping cloud-init tests on LXD $(lxd version) as \"cloud-init\" extension is not present"
+  # shellcheck disable=SC2034
+  FAIL=0
+  exit 0
+fi
+
+# Configure LXD
+lxd init --auto
+
+IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
+
+echo "==> Create key pairs for tests"
+ssh-keygen -t ed25519 -f "./profile-key" -N ""
+ssh-keygen -t ed25519 -f "./additional-key" -N ""
+
+echo "==> Create configure profile for tests"
+lxc profile set default cloud-init.user-data "$(cat <<EOF
+#cloud-config
+users:
+  - name: root
+    ssh_authorized_keys: $(cat ./profile-key.pub)
+  - name: ubuntu
+    ssh_authorized_keys: $(cat ./profile-key.pub)
+EOF
+)"
+
+echo "==> Create test instance"
+lxc init "${IMAGE}" c1
+
+if hasNeededAPIExtension "cloud_init_ssh_keys"; then
+  lxc config set c1 cloud-init.ssh-keys.mykey="root:$(cat ./additional-key.pub)"
+fi
+
+lxc start c1
+waitInstanceBooted c1
+
+echo "==> Test SSH after cloud-init seeded the authorized_keys files of the respective users"
+C_IPV4="$(lxc list c1 -c4 --format=csv | cut -d' ' -f1)"
+[ "$(ssh -o StrictHostKeyChecking=no -q -i ./profile-key "root@${C_IPV4}" whoami)" = "root" ]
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -89,6 +89,7 @@ jobs:
           - container
           - container-copy
           - conversion
+          - cloud-init
```
simondeziel, line 92: That list is sorted alphabetically ;)

tests/cloud-init
```diff
@@ -0,0 +1,38 @@
...
+echo "==> Create key pairs for tests"
+ssh-keygen -t ed25519 -f "./profile-key" -N ""
+ssh-keygen -t ed25519 -f "./additional-key" -N ""
+
+echo "==> Create configure profile for tests"
+lxc profile set default cloud-init.user-data "$(cat <<EOF
+#cloud-config
+users:
+  - name: root
+    ssh_authorized_keys: $(cat ./profile-key.pub)
```
simondeziel, line 21: Could you also inject a key in the `ubuntu` user's `authorized_keys` file too?

tests/cloud-init
```diff
@@ -0,0 +1,38 @@
...
+EOF
+)"
+
+echo "==> Create test instance"
+lxc launch "${IMAGE}" c -c cloud-init.ssh-keys.mykey="root:$(cat ./additional-key)"
+C_IPV4="$(lxc list c -c4 --format=csv | cut -d' ' -f1)"
+
+[ "$(ssh -o "StrictHostKeyChecking no" -q -i ./profile-key "root@${C_IPV4}" 'echo "ok"')" = "ok" ]
+[ "$(ssh -o "StrictHostKeyChecking no" -q -i ./additional-key "root@${C_IPV4}" 'echo "ok"')" = "ok" ]
```
simondeziel, line 30:
```suggestion
[ "$(ssh -o StrictHostKeyChecking=no -q -i ./profile-key "root@${C_IPV4}" whoami)" = "root" ]
[ "$(ssh -o StrictHostKeyChecking=no -q -i ./additional-key "root@${C_IPV4}" whoami)" = "root" ]
[ "$(ssh -o StrictHostKeyChecking=no -q -i ./profile-key "ubuntu@${C_IPV4}" whoami)" = "ubuntu" ]
! "$(ssh -o StrictHostKeyChecking=no -q -i ./additional-key "ubuntu@${C_IPV4}" whoami || false
```

tests/cloud-init
```diff
@@ -0,0 +1,38 @@
...
+lxc profile set default cloud-init.user-data "$(cat <<EOF
+#cloud-config
+users:
+  - name: root
+    ssh_authorized_keys: $(cat ./profile-key.pub)
+EOF
+)"
+
+echo "==> Create test instance"
+lxc launch "${IMAGE}" c -c cloud-init.ssh-keys.mykey="root:$(cat ./additional-key)"
```
simondeziel, line 26: To ease future test expansion, I'd probably name the container `c1`.

tests/cloud-init
```diff
@@ -0,0 +1,38 @@
...
+C_IPV4="$(lxc list c -c4 --format=csv | cut -d' ' -f1)"
+
```
simondeziel, line 28: I believe we'd need `waitInstanceBooted` here to let `cloud-init` finish its job.

simondeziel (commented):
Nice!

.github/workflows/tests.yml
hamistao: The worst part is that I knew and for some reason put `L` after `O` :woozy_face:

tests/cloud-init
```diff
@@ -0,0 +1,44 @@
...
+users:
+  - name: root
+    ssh_authorized_keys: $(cat ./profile-key.pub)
+  - name: ubuntu
+    ssh_authorized_keys: $(cat ./profile-key.pub)
+EOF
+)"
+
+echo "==> Create test instance"
+lxc launch "${IMAGE}" c1 -c cloud-init.ssh-keys.mykey="root:$(cat ./additional-key)"
```
simondeziel, line 28: Why provide the private key?

tests/cloud-init
```diff
@@ -0,0 +1,44 @@
...
+waitInstanceBooted c1
+
+echo "==> We can SSH using the injected keys's private counterparts"
```
simondeziel, line 31:
```suggestion
echo "==> Test SSH after cloud-init seeded the authorized_keys files of the respective users"
```
I'd like to avoid any potential confusion about private key injection (you said counterparts but still) as `cloud-init` should only be handed public keys, isn't it?

hamistao: Yes only public-keys should be sent to cloud-init. Will fix.

tests/cloud-init
hamistao: Whoops, I swallowed the `.pub`

tomponline (changes requested):
Failing on non latest/edge channels, so this isnt mergeable right now.

I suspect you need to do an API extension check for the SSH keys feature.

hamistao:
@tomponline This is fixed

tomponline (approved):
Thanks!

PR #413, opened 2025-02-13
tests/storage-volumes-vm: VM snapshot attachment

MggMuggins:
Snapshot attachment tests for https://github.com/canonical/lxd/pull/14930

`latest/edge` will fail until that PR is merged.

I'm still looking at a `lxc storage volume detach` bug with LVM instance snapshots, but it only affects detach when the instance's root volume is also mounted. I will put up a PR/modify this one once I've tracked it down.

### DIFF
--- a/tests/storage-volumes-vm
+++ b/tests/storage-volumes-vm
@@ -193,6 +193,8 @@ do
 
 	# attach VM root volumes
 	if hasNeededAPIExtension instance_root_volume_attachment; then
+		echo "==> Instance root & snapshot volumes"
+
 		empty_vm_size=8KiB
 		if [ "${poolDriver}" = "powerflex" ]; then
 			empty_vm_size=8GiB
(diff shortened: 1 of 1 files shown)


tests/storage-volumes-vm
```diff
@@ -272,6 +288,16 @@ do
 
 		lxc storage volume unset "${poolName}" virtual-machine/v2 security.shared
 
+		# Snapshots can be attached to profiles without security.shared because they
+		# will be mounted read-only
+		lxc profile create v2-snap0
+		lxc storage volume attach-profile "${poolName}" virtual-machine/v2/snap0 v2-snap0 v2-rs
+
+		lxc profile add v1 v2-snap0
```
simondeziel, line 296: Is it worth checking if the volume is hotplugged into `v1`?

simondeziel (commented):
LGTM on paper :)

tests/storage-volumes-vm
MggMuggins: Yes, I've found ~2 bugs since doing so.

simondeziel (commented):
The changes look good to me, thanks!

It's currently failing on the "Readonly custom block volumes are readonly" check you added but I assume that's related to the bug you found.

simondeziel (approved):

tomponline (approved):
thanks!

PR #415, opened 2025-02-14
Exclude `tests/cloud-init` on older releases

simondeziel:


simondeziel:
This assumes that `stable-5.21` will see that extension backported. I'm happy to exclude it as well if that's not planned.

tomponline:
> This assumes that `stable-5.21` will see that extension backported. I'm happy to exclude it as well if that's not planned.

yep leave it testing 5.21 for now as id like to backport.

tomponline (approved):

PR #424, opened 2025-03-11
tests/container: wait for snapd init before installing LXD snap

mihalicyn:
It is an attempt to fix failures like this:
```
+ echo '==> nested container (22.04)'

+ lxc launch ubuntu-minimal-daily:22.04 n1 -c security.nesting=true -c security.devlxd.images=true ==> nested container (22.04)
Launching n1

Retrieving image: Unpacking image: 100% (1.25GB/s)

Retrieving image: Unpacking image: 100% (1.25GB/s)
+ isSystemdClean n1
+ lxc exec n1 -- snap install lxd --channel=latest/edge error: cannot install "lxd": Post "https://api.snapcraft.io/v2/snaps/refresh":
       context canceled
+ cleanup
```

### DIFF
--- a/tests/container
+++ b/tests/container
@@ -143,7 +143,11 @@ for release in $RELEASES; do
 
     echo "==> nested container (${release})"
     lxc launch "${IMAGE}" n1 -c security.nesting=true -c security.devlxd.images=true
+    lxc exec n1 -- snap list || true
     isSystemdClean n1 || ignore_known_issues "n1" "nesting"
+    lxc exec n1 -- sh -c "$(declare -f waitSnapdSeed); waitSnapdSeed"
+    lxc exec n1 -- snap list || true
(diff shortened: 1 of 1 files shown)


mihalicyn:
Let's see if it actually helps before merging.

simondeziel (commented):
I agree that using `declare -f waitSnapdSeed` sounds like the way to go. Let's see how it goes in CI ;)

mihalicyn:
> I agree that using `declare -f waitSnapdSeed` sounds like the way to go. Let's see how it goes in CI

Unfortunately, it turned out that it's not enough. So I added a random `sleep 5` in between instance launch and `snap install lxd ...`. I have suspicion that something is going wrong on the GitHub side with runners as we have more random failures from today.

PR #427, opened 2025-03-12
tests/container: print more info before snap install

mihalicyn:
Let's print more info like IP address and also check if we can actually reach snapstore API.


tomponline:
ha, it passed now :(

tomponline (approved):

PR #438, opened 2025-03-31
images/voidlinux: Fix container shutdown

MusicDin:
Fix the VoidLinux shutdown by creating a service that ensures `/etc/runit/stopit` exists with permissiong `100` on each boot.

Passing build: https://github.com/MusicDin/lxd-ci/actions/runs/14172602615


simondeziel (approved):
Thanks!

PR #443, opened 2025-04-07
tests: Add `storage-metadata` suite for checking the improved format

roosterfish:
Second try after https://github.com/canonical/lxd-ci/pull/428 got merged too early.

Should be landed together with https://github.com/canonical/lxd/pull/15129.

The new tests storage-metadata performs instance and custom storage volume exports/imports as well as migrations between older and newer versions of LXD.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,6 +109,7 @@ jobs:
           - snapd
           - storage-buckets
           - storage-disks-vm
+          - storage-metadata
           - storage-vm btrfs
           - storage-vm ceph
           - storage-vm dir
@@ -145,6 +146,8 @@ jobs:
           - test: storage-buckets
             track: "4.0/candidate"
           - test: storage-disks-vm
+            track: "4.0/candidate"
+          - test: storage-metadata
             track: "4.0/candidate"
           - test: storage-vm btrfs
             track: "4.0/candidate"
@@ -187,6 +190,8 @@ jobs:
           - test: storage-buckets
             track: "4.0/edge"
           - test: storage-disks-vm
+            track: "4.0/edge"
+          - test: storage-metadata
             track: "4.0/edge"
           - test: storage-vm btrfs
             track: "4.0/edge"
--- a/tests/storage-metadata
+++ b/tests/storage-metadata
@@ -1,0 +1,342 @@
+#!/bin/bash
+set -eux
+
+# Checks the given backup file contents for format version 1.
+check_format_version1() {
+    backupFile="$1"
+    backupFileType="$2"
+
+    if [ "${backupFileType}" != "instance" ] && [ "${backupFileType}" != "volume" ]; then
+        echo "Invalid backup file type: ${backupFileType}"
+        exit 1
+    fi
+
+    # These fields are only relevant for instances.
+    if [ "${backupFileType}" = "instance" ]; then
+        [ "$(yq .config.container < "${backupFile}")" != "null" ]
+        [ "$(yq .config.pool < "${backupFile}")" != "null" ]
+        [ "$(yq '.config.snapshots | length' < "${backupFile}")" = "1" ]
+    fi
+
+    [ "$(yq .config.volume < "${backupFile}")" != "null" ]
+    [ "$(yq '.config.volume_snapshots | length' < "${backupFile}")" = "1" ]
+    [ "$(yq '.snapshots | length' < "${backupFile}")" = "1" ]
+
+    # New fields which should not appear in the old default format.
+    [ "$(yq .config.instance < "${backupFile}")" = "null" ]
+    [ "$(yq .config.version < "${backupFile}")" = "null" ]
+    [ "$(yq .config.pools < "${backupFile}")" = "null" ]
+    [ "$(yq .config.volumes < "${backupFile}")" = "null" ]
+}
+
+# Checks the given backup file contents for format version 2.
+check_format_version2() {
+    backupFile="$1"
+    backupFileType="$2"
+
+    if [ "${backupFileType}" != "instance" ] && [ "${backupFileType}" != "volume" ]; then
+        echo "Invalid backup file type: ${backupFileType}"
+        exit 1
+    fi
+
+    # These fields are only relevant for instances.
+    if [ "${backupFileType}" = "instance" ]; then
+        [ "$(yq .config.instance < "${backupFile}")" != "null" ]
+        [ "$(yq '.config.pools | length' < "${backupFile}")" = "1" ]
+    fi
+
+    [ "$(yq .config.version < "${backupFile}")" != "null" ]
+    [ "$(yq '.config.volumes | length' < "${backupFile}")" = "1" ]
+    [ "$(yq '.config.volumes.[0].snapshots | length' < "${backupFile}")" = "1" ]
+    [ "$(yq '.snapshots | length' < "${backupFile}")" = "1" ]
+
+    # Deprecated fields which should not appear in the new default format.
+    [ "$(yq .config.container < "${backupFile}")" = "null" ]
+    [ "$(yq .config.pool < "${backupFile}")" = "null" ]
+    [ "$(yq .config.volume < "${backupFile}")" = "null" ]
+    [ "$(yq .config.volume_snapshots < "${backupFile}")" = "null" ]
+}
+
+# Install dependencies.
+# Releases before 24.04 don't have a yq deb so use the snap instead.
+# shellcheck disable=SC1091
+. /etc/os-release
+if dpkg --compare-versions "${VERSION_ID}" ge 24.04; then
+    install_deps jq yq
+else
+    install_deps jq
+    waitSnapdSeed
+    snap install yq
+fi
+
+# Install LXD.
+install_lxd
+
+IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
+
+# Configure LXD.
+lxc network create lxdbr0
+lxc profile device add default eth0 nic network=lxdbr0
+lxc config set core.https_address :8443
+
+poolDriver="zfs"
+poolName="vmpool-${poolDriver}"
+
+echo "==> Create storage pool using driver ${poolDriver}"
+lxc storage create "${poolName}" "${poolDriver}"
+lxc profile device add default root disk pool="${poolName}" path=/
+
+# Do a matrix test and check the current version of LXD against latest, 5.21 and 5.0:
+# latest is using the new format both internally and by default for exports/metadata.
+# 5.21   is using the new format internally but falls back to the old format for exports/metadata.
+# 5.0    is using the old format both internally and for exports/metadata (same for 4.0).
+# New versions can be added to the matrix on demand.
+# The test suite knows which versions are using the old format version 1.
+# All other versions are tested against the new format version 2.
+for vers in "latest" "5.21" "5.0"; do
+    risk="$(echo "${LXD_SNAP_CHANNEL}" | cut -d/ -f2)"
+    versStr="${vers}/${risk}"
+
+    # Don't test 5.0 clients against latest track as the client is too old and only supports password authentication.
+    # Due to the nature of this matrix test, the latest client gets tested against an 5.0 server.
+    if ! echo "${LXD_SNAP_CHANNEL}" | grep -qE "^(latest|5\.[1-9].)/" && [ "${vers}" = "latest" ]; then
+        echo "Skip testing ${LXD_SNAP_CHANNEL} against ${versStr} as the client is too old"
+        continue
+    fi
+
+    echo "==> Create container with a ${versStr} LXD"
+    lxc launch "${IMAGE}" c1 -s "${poolName}" -c limits.cpu=4 -c limits.memory=4GiB -d root,size=20GiB
+    waitInstanceReady c1
+
+    echo "==> Add required devices for creating empty VMs inside of the container"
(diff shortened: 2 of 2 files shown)


simondeziel (commented):
If that's the exact same code as before, `git revert`'ing my reverts might be cleaner.
Otherwise, LGTM.

roosterfish:
> If that's the exact same code as before, git revert'ing my reverts might be cleaner.

Good point, better for traceability, I have replaced the commits with the reverts.

roosterfish:
The test on `latest/edge` is working fine. 
The `5.21` track test is skipping as the new format isn't yet backported (see https://github.com/canonical/lxd-ci/actions/runs/14596657945/job/40944331502?pr=443#step:12:112).

Instead of skipping with `exit 0` I could do `exit 1` so that it fails for `5.21/edge` but it will go green after the changes are backported.

tomponline:
> Instead of skipping with `exit 0` I could do `exit 1` so that it fails for `5.21/edge` but it will go green after the changes are backported.

No thats fine thanks, as long as we're testing the old behaviour on 5.21/edge until it gets backported thats fine.

tests/storage-metadata
```diff
@@ -0,0 +1,262 @@
...
+# Workaround for https://bugs.launchpad.net/snapd/+bug/2104066
+[Service]
+Environment=SNAPD_STANDBY_WAIT=1m
+EOF
+    lxc exec v1 -- systemctl daemon-reload
+    lxc exec v1 -- systemctl restart snapd.service
+
+    # shellcheck disable=SC3044 # Ignore "declare is undefined" shellcheck error.
+    lxc exec v1 -- sh -c "$(declare -f waitSnapdSeed); waitSnapdSeed"
+    lxc exec v1 -- snap install lxd --channel="${vers}/edge"
```
tomponline, line 58: hrm should we be hardcoding `edge` here? What about when i run the test suite for `*/candidate` pre-release?

tests/storage-metadata
```diff
@@ -0,0 +1,262 @@
...
+    FAIL=0
+    exit 0
+fi
+
+# Do a matrix test and check both LXD 5.0 and 5.21
+# 5.0 is using the old format both internally and for exports.
+# 5.21 is already using the new format internally but falls back to the old format for exports.
+for vers in "5.0" "5.21"; do
+    echo "==> Create VM with a ${vers} LXD server using the original (old) backup format"
+    lxc launch "${IMAGE}" v1 --vm -s "${poolName}" -c limits.cpu=4 -c limits.memory=4GiB -d root,size=20GiB
```
tomponline, line 43: can we use a container for the outer LXD?

.github/workflows/tests.yml
```diff
@@ -227,6 +232,10 @@ jobs:
             track: "5.0/candidate"
           - test: vm-migration
             track: "5.0/edge"
+          - test: storage-metadata
```
tomponline, line 235: why is this not compatible, dont we test the old v1 export format too?

tests/storage-metadata
```diff
@@ -0,0 +1,262 @@
...
+    echo "LXD version $(lxd version | awk '{print $1}') doesn't support the new backup format, skipping test"
+    # shellcheck disable=SC2034
+    FAIL=0
+    exit 0
+fi
+
+# Do a matrix test and check both LXD 5.0 and 5.21
+# 5.0 is using the old format both internally and for exports.
+# 5.21 is already using the new format internally but falls back to the old format for exports.
+for vers in "5.0" "5.21"; do
```
tomponline, line 41: please can you clarify this comment, as the GH workflow itself is defining the versions of LXD to use in its matrix, so what is this one doing?

.github/workflows/tests.yml
roosterfish: The test suite is written with the intention to run on all versions (5.21, 6.x) which know about the new format. Then throughout the test an old (5.0) LXD is deployed which is used for export/import and migration from old to new/new to old format.
This ensures there is always an old LXD which doesn't know about backup metadata versions.

But I see we would gain much more test coverage by not excluding it as we could still check for the old format in those cases. I'll push some changes.

tests/storage-metadata
roosterfish: I tried to explain it in [this comment](https://github.com/canonical/lxd-ci/pull/443#discussion_r2054276597).
Having another matrix inside the test suite allows testing cross version support based on https://docs.google.com/spreadsheets/d/1hVr-ceuRbYY1RSYrzqRfxlOzS1VDAwq8csqM05R7ih0/edit?gid=0#gid=0

tests/storage-metadata
simondeziel: Initially I thought it would not be convenient because VMs are created onto this `v1` instance. After looking more closely, the instances on top of `v1` are never started/launched so I think that'd work.

.github/workflows/tests.yml
```diff
@@ -109,6 +109,7 @@ jobs:
           - snapd
           - storage-buckets
           - storage-disks-vm
+          - storage-metadata
```
simondeziel, line 112: Testing against all LXD versions turned up at least an interesting issue: https://github.com/canonical/lxd-ci/actions/runs/14619704868/job/41016390079?pr=443#step:12:377

roosterfish: Mh, or let's do `lxc exec v1 -- lxc export i1 && lxc file pull v1/i1.tar.gz .`? This way we can at least properly run the test.

tests/storage-metadata
roosterfish: I tried earlier and it required to install some zfs utils inside the container so we can create the storage pool using zfs driver. 
Should I rather use the `snap set lxd zfs.external=true`?

roosterfish: With `ubuntu-minimal-daily:24.04` container image I got at least an error from LXD complaining about missing tooling.

tomponline: what LXD version were you installing?

tomponline: sounds like it was a kernel mismatch

simondeziel: If you are referring to LXD 4.0 on Ubuntu 22.04 runner (like here https://github.com/canonical/lxd-ci/actions/runs/14619704868/job/41016409936?pr=443#step:12:102) indeed the problem is because 4.0 doesn't support 24.04's 6.8 kernel

tomponline: I think we can skip LXD 4.0 for this test

roosterfish: The error is this `Error: The requested backend 'zfs' isn't available on your system (missing tools)` when running `lxd init --auto --storage-backend=zfs '--network-address=[::]' --network-port=8443`.

Host kernel is `6.8.0-57-generic`, the container image is `ubuntu-minimal-daily:24.04` and the version of LXD I have installed in the container is `5.21/edge`.

tomponline: do we need to use ZFS inside the guest at all?

roosterfish: No, we can also do `dir` as the volumes are anyway only 1MiB in size. 
However when trying to create empty VM's I now get `Error: Failed creating instance record: Instance type "virtual-machine" is not supported on this server: KVM support is missing (no /dev/kvm)`.

In the VM nesting test I saw adding those devices to the container, but do we want to do the same here or is it not worth the overhead as we can just run with a VM?

tomponline: the overhead of running the VM is worth avoiding if we just need to pass through a couple of disk passthrough devices in my view

simondeziel: The whole test completes in 5m30s in CI so I'd personally vote for simplicity and stick to a VM here.

roosterfish: A little bit, I already moved it to a container so we can do it :)

>  after you worked on the tests for nesting previously

Cannot remember this, but I remember doing some matrix tests for migration/copy but without container/VM nesting.

.github/workflows/tests.yml
roosterfish: Thanks for the suggestion, I remember seeing this in another test.

But as the test also does migration between the two LXD servers we require the remote so I would say we stick with using a remote but skip 5.0 -> latest/edge. 
In any case (as its a matrix test), the test will run the other way around. Think what I will do is just run the migration in both directions.

tests/storage-metadata
roosterfish: Fixed, its now using a container with the necessary devices to init VMs.

.github/workflows/tests.yml
roosterfish: Fixed. Only 4.0 is excluded now.

tests/storage-metadata
roosterfish: Put some more comments, also how to extend this test in the future. Fixed.

.github/workflows/tests.yml
roosterfish: Fixed.

tests/storage-metadata
```diff
@@ -0,0 +1,342 @@
...
+
+    # Deprecated fields which should not appear in the new default format.
+    [ "$(yq .config.container < "${backupFile}")" = "null" ]
+    [ "$(yq .config.pool < "${backupFile}")" = "null" ]
+    [ "$(yq .config.volume < "${backupFile}")" = "null" ]
+    [ "$(yq .config.volume_snapshots < "${backupFile}")" = "null" ]
+}
+
+# Install dependencies.
+# Releases before 24.04 don't have a yq deb so use the snap instead.
```
tomponline, line 61: just a nit, why dont we just always use the snap for consistency?

tests/storage-metadata
```diff
@@ -0,0 +1,342 @@
...
+            # The old format should be allowed for import in any case.
+            echo "==> Check the ${type} can be imported on LXD ${LXD_SNAP_CHANNEL}"
+            lxc import "${tmpdir}/i1.tar.gz" i1
+            [ "$(lxc query "/1.0/instances/i1?recursion=1" | jq '.snapshots | length')" = "1" ]
+
+            lxc delete -f i1
+        else
+            check_format_version2 "${tmpdir}/backup/index.yaml" instance
+
+            # Only LXD >= 5.21 allows importing the new format.
```
tomponline, line 170: wont this comment be out of date once we backport support for v2 format to 5.21?

tests/storage-metadata
```diff
@@ -0,0 +1,342 @@
...
+                [ "$(lxc query "/1.0/instances/i1?recursion=1" | jq '.snapshots | length')" = "1" ]
+
+                lxc delete -f i1
+            fi
+        fi
+
+        rm -rf "${tmpdir}/backup"
+        rm "${tmpdir}/i1.tar.gz"
+
+        # Only LXD >= 5.21 allows setting the export version.
```
tomponline, line 183: wont this comment be out of date once we backport support for v2 format to 5.21?

tests/storage-metadata
```diff
@@ -0,0 +1,342 @@
...
+            lxc export c1:i1 "${tmpdir}/i1.tar.gz" --export-version 2
+            tar -xzf "${tmpdir}/i1.tar.gz" -C "${tmpdir}" backup/index.yaml
+
+            echo "==> Check the backup index file"
+            cat "${tmpdir}/backup/index.yaml"
+            check_format_version2 "${tmpdir}/backup/index.yaml" instance
+
+            rm -rf "${tmpdir}/backup"
+
+            # Only LXD >= 5.21 allows importing the new format 2.
```
tomponline, line 214: wont this comment be out of date once we backport support for v2 format to 5.21?

tests/storage-metadata
roosterfish: If a proactively added this check `if hasNeededAPIExtension backup_metadata_version;` so that it skips 5.21 for now as it doesn't yet have the `backup_metadata_version` extension. 
But once its backported, all versions starting from 5.21 understand the new format which is what the comment describes.
Hope that's clear but I can change the sentence if you want.

tomponline: Oh i see, i missed the `=` of `>=`. Got you.

tests/storage-metadata
roosterfish: The valid argument is that the apt package is faster to install, see also this thread from the old PR https://github.com/canonical/lxd-ci/pull/428#discussion_r1994211944.

tests/storage-metadata
```diff
@@ -0,0 +1,342 @@
...
+# The test suite knows which versions are using the old format version 1.
+# All other versions are tested against the new format version 2.
+for vers in "latest" "5.21" "5.0"; do
+    risk="$(echo "${LXD_SNAP_CHANNEL}" | cut -d/ -f2)"
+    versStr="${vers}/${risk}"
+
+    # Don't test 5.0 clients against latest track as the client is too old and only supports password authentication.
+    # Due to the nature of this matrix test, the latest client gets tested against an 5.0 server.
+    if ! echo "${LXD_SNAP_CHANNEL}" | grep -qE "^(latest|5\.[1-9].)/" && [ "${vers}" = "latest" ]; then
+        echo "Skip testing ${LXD_SNAP_CHANNEL} against ${versStr} as the client is too old"
```
simondeziel, line 103: Could we use `hasNeededAPIExtension` here to detect if the client side doesn't have what's needed to interact with `${vers}`?

roosterfish: Mh, it seems we cannot easily tell this based on present API extensions. The `explicit_trust_token` extension indicates the server supports using the token field explicitly. 
But this wouldn't help to identify if support for the legacy password got dropped (as we did starting from 6.x/latest).

PR #447, opened 2025-04-24
tests/cluster: Improve `fan` connectivity tests

simondeziel:
The existing `tests/cluster` script didn't catch https://github.com/canonical/lxd/issues/15212 because upgrading LXD to the broken revision didn't touch the already created `lxdfan0` network device and the inner instances were not checked for connectivity after the upgrade anyway. https://github.com/canonical/lxd-ci/actions/runs/13917046846/job/38941869251 was one of the last test runs with the broken LXD revision.

Reboot the whole cluster after it's been upgraded has shown to catch the problem after I manually side-loaded a broken LXD (main with https://github.com/canonical/lxd/pull/15213/commits/197a0238f7f7d80c123f6e0c1716f4efc0e8de2f reverted).

The rest of the PR tries to reduce the runtime so make up for the delay introduced by the cluster reboot. In terms of speed, the end result is ~20-30s faster in GH CI but a tad slower locally.

### DIFF
--- a/tests/cluster
+++ b/tests/cluster
@@ -5,7 +5,7 @@ set -eux
 install_lxd
 
 # Configure LXD
-lxd init --auto
+lxd init --auto --storage-backend=btrfs
 
 IMAGE="${TEST_IMG:-ubuntu-minimal-daily:24.04}"
 
--- a/tests/cluster
+++ b/tests/cluster
@@ -34,12 +34,14 @@ instance_connectivity_checks() {
 }
 
 wait_for_online_cluster() {
-    for _ in $(seq 10); do
-        sleep 30
+    sleep 5
+    lxc exec m1 -- lxd waitready --timeout=300
+    for _ in $(seq 30); do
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Great thanks!

tomponline:
> > @simondeziel does the main lxd test suite test fan networking btw?
> 
> No and that was a conscientious decision:
> 
> ```
> test_clustering_dns() {
>   local lxdDir
> 
>   # Because we do not want tests to only run on Ubuntu (due to cluster's fan network dependency)
>   # instead we will just spawn forkdns directly and check DNS resolution.
> ```

We should probably change/add ubuntu specific fan tests so we can catch regressions earlier and contribute to coverage tests.

simondeziel:
> We should probably change/add ubuntu specific fan tests so we can catch regressions earlier and contribute to coverage tests.

Sounds good. I will add logic to detect FAN support and fail hard if not found and the host OS is Ubuntu.

PR #469, opened 2025-05-07
images/openwrt: Remove arch condition

MusicDin:
Fixes https://github.com/canonical/lxd/issues/15321


MusicDin:
The architecture condition was never met. Mapped architecture for openWRT is `x86_64` instead of `amd64`. However, we want to enable console access on arm based images as well, therefore the condition is entirely removed. Inaccessible console on arm images was confirmed by @mihalicyn

tomponline (approved):
Thanks both!

PR #471, opened 2025-05-18
github: don't build/publish images during the weekends

simondeziel:
Most distributions do not publish images during the weekends for obvious reasons so rebuilding/republishing those works against the local cache that LXD manages.

Closes https://github.com/canonical/lxd/issues/13294


MusicDin (approved):
Thanks :rocket:

PR #473, opened 2025-05-19
github/testflinger: use stable channel for pc-kernel snap

mihalicyn:
Now we can (and should) use 24/stable channel for pc-kernel snap instead of temporary and expired 24/edge/nvidia-components-dev.

Tested with:
JOB_QUEUE=lxd-nvidia SNAP_CHANNEL=latest/edge DISTRO=core24-latest ./run.sh

from local machine.


simondeziel (approved):
Thanks

PR #476, opened 2025-05-21
tests/storage-disks-vm: fix debug logic

simondeziel:
Provide the right instance name to `lxc info --show-log` and avoid the subshell.


simondeziel:
I think it's the first time all 178 checks are green on the first try... today's my lucky day ;)

PR #480, opened 2025-05-23
Update Mint images

simondeziel:


simondeziel:
Built fine on my fork: https://github.com/simondeziel/lxd-ci/actions/runs/15214220301/job/42795667385

MusicDin (approved):
Thanks! Nice idea for EOL comments :)

simondeziel:
Speaking of Mint, I was wondering what was the point of shipping non-desktop images of Mint since AFAIK, Mint is a flavour mostly for its different desktop environment. I think we should think about publishing desktop images for it.

PR #491, opened 2025-05-30
tests/container: stop testing 20.04 instances (EOL)

simondeziel:


escabo:
> @simondeziel @escabo shall we close this for now?

Yes.

PR #502, opened 2025-06-10
tests/network-routed: test both container and VM

simondeziel:


simondeziel:
@tomponline now that we test both containers and VMs, the `latest/edge` fails where we expect it to fail: when `cloud-init` couldn't configure the needed network bits. I added additional debug info that further confirm the network is not configured in the failing cases.

tomponline (approved):

PR #525, opened 2025-07-21
tests: Update devlxd-container to check instance name validation methods

tomponline:


tomponline:
devlxd-container failures on channels less than latest/edge are expected.

tomponline:
All expected devlxd-container tests have passed.

tests/devlxd-container
```diff
@@ -33,6 +33,23 @@ lxc exec c1 -- curl --silent --show-error --unix-socket /dev/lxd/sock http://cus
 lxc exec c1 -- curl --silent --show-error --unix-socket /dev/lxd/sock http://custom.socket/1.0/meta-data | grep -Fq 'instance-id:'
 lxc exec c1 -- curl --silent --show-error --unix-socket /dev/lxd/sock http://custom.socket/1.0/meta-data | grep -Fq 'local-hostname: c1'
 
+# Check devlxd instance name validation works (5.21 onwards).
+if ! echo "${LXD_SNAP_CHANNEL}" | grep -qE "^[45]\.0/"; then
+  lxc launch "${TEST_IMG:-ubuntu:24.04}" c2 -s "${poolName}"
+
+  # First check first discovery method (where curl process isn't under the lxc monitor process tree).
+  lxc exec c1 -- bash -c "exec -a '[lxc monitor]' curl --silent --show-error --unix-socket /dev/lxd/sock http://custom.socket/1.0/meta-data --proxy ' default_c2'" | grep -Fq 'local-hostname: c1'
```
roosterfish, line 41: I have now looked at this for quite some time and don't get what the `--proxy` param is used for. The format looks like `project_instanceName` but what is the effect of this? Especially with the added whitespace? Setting a proxy for any request is clear but there seems to be more to this.

This might seem obvious but I never ran across this.

simondeziel: It seems like the exec on `c1` is trying to masquerade as a legitimate request supposedly originating from `c2` itself and trying to get its meta-data.

tomponline: I will update with a comment.

roosterfish: Ah thanks, nifty solution :)

roosterfish (approved):

PR #526, opened 2025-07-22
tests: Check that volume is functional after live migration

MusicDin:
Run on testflinger instead of GH runner to lower 1 level of nested virtualization.

Based on: https://github.com/canonical/lxd-ci/pull/256


.github/workflows/vm-migration.yml
```diff
@@ -0,0 +1,69 @@
...
+          - '["5.21/stable"]'
+
+jobs:
+  vm-migration:
+    runs-on: [self-hosted, self-hosted-linux-amd64-jammy-private-endpoint-medium]
+    strategy:
+      matrix:
+        track: ${{ fromJSON(inputs.snap-track || '["latest/edge"]') }}
+        os:
+          - core24-latest
```
simondeziel, line 43: I think this means to test on Ubuntu Core which is potentially not what we want here.

PR #533, opened 2025-07-23
tests: Run container nesting tests on 5.0/edge

MusicDin:
WIP

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -142,6 +142,11 @@ jobs:
           - vm
           - vm-migration
           - vm-nesting
+        include:
+          # Run container-nesting tests on 24.04 as well for comparison.
+          - test: container-nesting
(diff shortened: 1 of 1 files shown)


simondeziel:
@mihalicyn any idea what could cause `ERROR    mount_utils - ../src/src/lxc/mount_utils.c:move_detached_mount:326 - Permission denied - Failed to attach detached mount 20 to filesystem at 18` here?

simondeziel:
Hmm, maybe the root case has more to do with the first error: `ERROR    cgroup2_devices - ../src/src/lxc/cgroups/cgroup2_devices.c:bpf_program_load_kernel:332 - Operation not permitted - Failed to load bpf program: (null)`

simondeziel:
Found the mundane reason (see #538)... definitely not due to the kernel :/

PR #539, opened 2025-07-25
tests/container: also test Ubuntu 25.04/25.10

simondeziel:


simondeziel:
I'll revisit this when the time permits.

tomponline:
@simondeziel did it not work?

PR #540, opened 2025-07-25
LXD UI tweaks

simondeziel:


simondeziel:
@edlerd I've made the changes I mentioned in https://github.com/canonical/lxd-ci/pull/527#pullrequestreview-3048317643.

I am not sure if those changes could explain the breakage I see here but looking at the video captures, I couldn't pin point what went wrong. Your input would be appreciated, please.

PR #541, opened 2025-07-28
github: tweak image building/publication schedule

simondeziel:
This change does two things:

* Spread out the image building/publication to avoid overloading `images:`
 * Build `arm64` images daily now that `arm64` runners are free
   (addresses parts of https://github.com/canonical/lxd/issues/15651)

This should help with the SFTP upload we have especially on Sundays because `arm64` and `amd64` builds were triggered at the exact same time for all distros.

### DIFF
--- a/.github/actions/image-upload/action.yml
+++ b/.github/actions/image-upload/action.yml
@@ -59,8 +59,14 @@ runs:
         # First upload contents to the temporary dir and once fully uploaded
         # move the directory to the final destination to avoid potential race
         # where simplestream-maintainer includes partially uploaded images.
-        sftp -oHostKeyAlgorithms=ssh-ed25519 -P ${SSH_PORT} -oConnectionAttempts=10 -oConnectTimeout=10 -b - "${SSH_USER}@${SSH_HOST}" <<EOF
-            put -R "${SRC_DIR}"-upload/*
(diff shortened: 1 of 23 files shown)


tomponline (approved):
Grudgingly approving, but having jump through these hoops to upload some files by sftp seems to point at a problem on the new server/platform.

simondeziel:
> Grudgingly approving, but having to jump through these hoops to upload some files by sftp seems to point at a problem on the new server/platform.

While I understand your point I think it's still worthwhile to spread out the building/uploading of those images. Maybe you'd like me to "more densely" pack them? In other words try to spread them over say 3 hours instead of spacing them by 15 minutes per image/arch?

tomponline:
> > Grudgingly approving, but having to jump through these hoops to upload some files by sftp seems to point at a problem on the new server/platform.
> 
> While I understand your point I think it's still worthwhile to spread out the building/uploading of those images. Maybe you'd like me to "more densely" pack them? In other words try to spread them over say 3 hours instead of spacing them by 15 minutes per image/arch?

I don't mind the spreading you've done here its fine, I'm just concerned its being required because of some underlying connectivity/rate limiting issue that we (Canonical) should identify and resolve in case it's also affecting downloading of images too.

PR #546, opened 2025-07-30
github: replace NixOS 24.05 (EOL) by newer 25.05

simondeziel:


tomponline (approved):

adamcstephens:
FYI, we do two releases a year, with a month overlap. So I'd recommend not skipping the November release. :)

PR #547, opened 2025-07-31
test/storage-disks-vm: Update tests to check for directory disk hotplug support during guest boot

tomponline:
As the guest-boot time hotplug mount and the device removal unmount changes are bug fixes they don't have their own API extensions, so ive used the `io.threads` section as any LXD that has support for that should also have these bug fixes.


simondeziel (approved):
Thanks!

PR #550, opened 2025-07-31
tests: Check that load balancers and forwards are reachable from inside the OVN network

nmezhenskyi:
Follow-up to https://github.com/canonical/lxd/pull/16119.

Add tests to check that load balancers and forwards are reachable from inside the ovn network.

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -779,12 +779,18 @@ ovn_forward_tests() {
 interface=eth0
 bind-interfaces
 interface-name=u1.lxd,eth0
+host-record=lxd.localdomain,127.0.0.1
+host-record=lxd.localdomain,::1
 EOF
 
     lxc exec u1 -- systemctl unmask dnsmasq
     lxc exec u1 -- systemctl start dnsmasq
     dig a +tcp @192.0.2.1 u1.lxd
     dig aaaa +tcp @2001:db8:1:2::1 u1.lxd
+
+    echo "==> Check accessing forwards's listen address inside the u1"
+    [ "$(lxc exec u1 -- dig a +tcp @192.0.2.1 lxd.localdomain +short)" = "127.0.0.1" ]
+    [ "$(lxc exec u1 -- dig aaaa +tcp @2001:db8:1:2::1 lxd.localdomain +short)" = "::1" ]
 
     echo "==> Check stopping the DNS service and checking we get no connection using default target rule"
     lxc exec u1 -- systemctl stop dnsmasq
@@ -1044,6 +1050,7 @@ interface=eth0
 bind-interfaces
 interface-name=u1.lxd,eth0
 host-record=lxd.localdomain,127.0.0.1
+host-record=lxd.localdomain,::1
 EOF
 
     lxc exec u1 -- systemctl unmask dnsmasq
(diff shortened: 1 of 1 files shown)


tests/network-ovn
```diff
@@ -1051,6 +1051,10 @@ EOF
     dig a +tcp @192.0.2.1 u1.lxd
     dig aaaa +tcp @2001:db8:1:2::1 u1.lxd
 
+    echo "==> Check accessing load balancer's listen address inside the u1"
+    lxc exec u1 -- dig a +tcp @192.0.2.1 u1.lxd
+    lxc exec u1 -- dig aaaa +tcp @2001:db8:1:2::1 u1.lxd
```
simondeziel, line 1056: We should also check that stopping `dnsmasq` cause those lookups to fail.

tests/network-ovn
```diff
@@ -1051,6 +1051,10 @@ EOF
     dig a +tcp @192.0.2.1 u1.lxd
     dig aaaa +tcp @2001:db8:1:2::1 u1.lxd
 
+    echo "==> Check accessing load balancer's listen address inside the u1"
+    lxc exec u1 -- dig a +tcp @192.0.2.1 u1.lxd
```
simondeziel, line 1055: If I understand the test right, we should not be looking up the instance FQDN but `lxd.localdomain` that is the synthetic record that `dnsmasq` should return.

Ideally, we'd do a `dig +short` and compare the output with the expected `127.0.0.X`.

I know you just copied what was done before but I fear the risk of accidentally poking at LXD/OVN default DNS responder.

tests/network-ovn
```diff
@@ -1051,11 +1051,17 @@ EOF
     dig a +tcp @192.0.2.1 u1.lxd
     dig aaaa +tcp @2001:db8:1:2::1 u1.lxd
 
+    echo "==> Check accessing load balancer's listen address inside the u1"
+    lxc exec u1 -- dig a +tcp @192.0.2.1 lxd.localdomain +short
```
simondeziel, line 1055: Sorry for not being clear here's what I had in mind:
```suggestion
    [ "$(lxc exec u1 -- dig a +tcp @192.0.2.1 lxd.localdomain +short)" = "127.0.0.1" ]
```

tests/network-ovn
```diff
@@ -1091,6 +1097,10 @@ EOF
     dig a +tcp @192.0.2.1 u2.lxd
     dig aaaa +tcp @2001:db8:1:2::1 u2.lxd
 
+    echo "==> Check accessing load balancer's listen address inside the u2"
+    lxc exec u2 -- dig a +tcp @192.0.2.1 lxd.localdomain +short
```
simondeziel, line 1101: [ "$(lxc exec u1 -- dig a +tcp @192.0.2.1 lxd.localdomain +short)" = "127.0.0.2" ]

simondeziel: Oh my bad, I assumed IPv6 was used only for the DNS server address but you are right, it's making a `AAAA` request. Adding `AAAA` synthetic records would work. Thanks :)

simondeziel (approved):
LGTM, thanks!
Tests are all green with `latest/edge` but not with older channels so that's good.

tests/network-ovn
tomponline: The AAAA request isn't strictly needed here, we could have done an A request over ipv6 to test the ipv6 load balancer was working. But either approach is fine.

tomponline (approved):
LGTM

PR #553, opened 2025-08-01
Fix 2 issues with `arm64` images

simondeziel:
Slackware uses a separate site to distribute `arm64` builds so it would need
more work on lxd-imagebuilder.


simondeziel:
The gentoo `arm64` images built fine on my work with the change.

PR #554, opened 2025-08-01
tests: Test ZFS in nested containers

MusicDin:
Test ZFS in nested containers using ZFS delegation.
This also required ZFS version >= `2.2`.

Note that `/dev/zfs` permissions are too strict, and we have a workaround in place to allow usage of ZFS inside the nested container.

### DIFF
--- a/tests/container-nesting
+++ b/tests/container-nesting
@@ -59,6 +59,15 @@ for poolDriver in ${poolDriverList}; do
         if dpkg --compare-versions "${zfsVersion}" gt "2.2"; then
             lxc storage set "${poolName}" volume.zfs.delegate=true
             nestedPoolDriverList+=" zfs"
+
+            # XXX: Ensure that the ZFS device is accessible within the nested container.
+            #      Otherwise, the LXD storage pool creation will fail.
+            zfsPerm=$(stat -c '%a' /dev/zfs)
+            if [ $((zfsPerm & 7)) -eq 0 ]; then
+                echo "::warning::Workaround for ZFS permission is still in effect."
+                ls -la /dev/zfs
+                chmod 666 /dev/zfs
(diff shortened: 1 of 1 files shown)


tests/container-nesting
```diff
@@ -52,6 +52,25 @@ for poolDriver in ${poolDriverList}; do
         # BTRFS can be used within a nested container on top of BTRFS storage
         # pool, if the source is a directory.
         nestedPoolDriverList+=" btrfs"
+    elif [ "${poolDriver}" = "zfs" ]; then
+        # To use ZFS within a nested container, a ZFS version must be greater than
+        # or equal to 2.2, and ZFS delegation must be enabled.
+        zfsVersion=$(modinfo zfs | awk '/^version:/ {print $2}' | cut -d- -f1)
```
simondeziel, line 58: There is no need to `cut` anything. `dpkg --compare-versions` has the smart needed to cope with `2.2.2-0ubuntu9.2`.

tests/container-nesting
```diff
@@ -97,6 +121,9 @@ for poolDriver in ${poolDriverList}; do
 
     # Cleanup.
     lxc delete --force "${instName}"
+    if echo "${nestedPoolDriverList}" | grep -q zfs; then
```
simondeziel, line 124: When matching in lists, it's best to use `-w` to avoid matching on substrings. Granted, the risk of having another driver with `zfs` in it's name is very low.
```suggestion
    if echo "${nestedPoolDriverList}" | grep -qwF zfs; then
```

simondeziel (approved):

tests/container-nesting
MusicDin: Thanks!

PR #560, opened 2025-08-11
tests: Add tests for internal load balancers and forwards

nmezhenskyi:
This PR adds tests to check that internal load balancers and forwards are reachable from inside of the OVN network, but are unreachable from outside.

Follow-up to https://github.com/canonical/lxd/pull/16162.

This should be merged together with https://github.com/canonical/lxd/pull/16179 (adds `ovn_internal_load_balancer` API extension for LXD).

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -876,6 +876,23 @@ EOF
     lxc network delete ovn-virtual-network2
     [ "$(ovn-nbctl list load_balancer | grep -cF name)" = 4 ]
 
+    if hasNeededAPIExtension ovn_internal_load_balancer; then
+        echo "==> Test internal forwards"
+
+        echo "==> Check creating internal forwards"
+        lxc network forward create ovn-virtual-network "${ip4AddressPrefix}.10" target_address="${U1_IPV4}"
+        lxc network forward create ovn-virtual-network "${ip6AddressPrefix}::10" target_address="${U1_IPV6}"
+        [ "$(ovn-nbctl list load_balancer | grep -cF name)" = 6 ]
+
+        echo "==> Check we get no connection outside of the OVN network"
+        ! dig a +tcp "@${ip4AddressPrefix}.10" u1.lxd || false
+        ! dig aaaa +tcp "@${ip6AddressPrefix}::10" u1.lxd || false
+
+        echo "==> Check accessing forwards's listen address inside the u1"
+        [ "$(lxc exec u1 -- dig a +tcp "@${ip4AddressPrefix}.10" lxd.localdomain +short)" = "127.0.0.1" ]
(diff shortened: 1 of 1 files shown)


tests/network-ovn
```diff
@@ -1144,6 +1161,57 @@ EOF
...
+        lxc network load-balancer backend add ovn-virtual-network "${ip4AddressPrefix}.10" u1 "${U1_IPV4}"
+        lxc network load-balancer backend add ovn-virtual-network "${ip6AddressPrefix}::10" u1 "${U1_IPV6}"
+
+        echo "==> Add backend for u2"
+        lxc network load-balancer backend add ovn-virtual-network "${ip4AddressPrefix}.10" u2 "${U2_IPV4}"
+        lxc network load-balancer backend add ovn-virtual-network "${ip6AddressPrefix}::10" u2 "${U2_IPV6}"
+
+        echo "==> Configure ports for the load balancers towards u1 and u2"
+        lxc network load-balancer port add ovn-virtual-network "${ip4AddressPrefix}.10" tcp 53 u1,u2
+        lxc network load-balancer port add ovn-virtual-network "${ip6AddressPrefix}::10" tcp 53 u1,u2
```
simondeziel, line 1185: I'm not up to date on what LXD/OVN support so is this load-balancing taking into account the health of the targets? In other words, if we `lxc stop u1`, will `u2` pick up the slack?

nmezhenskyi: > is this load-balancing taking into account the health of the targets?

No, it does not. If we stop `u1` and the request gets sent there, it will just hang. We have a separate roadmap item for health checks that will prevent this, but it's not a part of this cycle AFAIK.

tests/network-ovn
```diff
@@ -1144,6 +1161,57 @@ EOF
...
+        ovn-nbctl list load_balancer | grep -F "{\"[${ip6AddressPrefix}::10]:53\"=\"[${U1_IPV6}]:53,[${U2_IPV6}]:53\"}"
+
+        echo "==> Check DNS load balancing is working and is accessible from u1 and u2, but not from outside of the OVN network"
+        for _ in $(seq 5); do
+            ! dig a +tcp "@${ip4AddressPrefix}.10" lxd.localdomain +short || false
+            ! dig aaaa +tcp "@${ip6AddressPrefix}::10" lxd.localdomain +short || false
+            lxc exec u1 -- dig a +tcp "@${ip4AddressPrefix}.10" lxd.localdomain +short
+            lxc exec u1 -- dig aaaa +tcp "@${ip6AddressPrefix}::10" lxd.localdomain +short
+            lxc exec u2 -- dig a +tcp "@${ip4AddressPrefix}.10" lxd.localdomain +short
+            lxc exec u2 -- dig aaaa +tcp "@${ip6AddressPrefix}::10" lxd.localdomain +short
```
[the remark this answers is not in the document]
nmezhenskyi: Hm, this should work

PR #562, opened 2025-08-13
github: do SFTP connectivity test in `actions/image-setup`

simondeziel:
A `tcpdump` capture on the PS6 VM showed that failed upload were accompanied with exactly 0 traffic from GH reaching the PS6 VM. Since this traffic is subject to IP allow-lists, my current theory is that IS' list does not include an up to date list of GH possible IPv4 source addresses.

As such, it's best to test the connectivity before doing anything "expensive" with the GH runner. Also, when the connectivity will fail again, I'll be armed with the public IPv4 that was used for the upload test and this will let me correlate it with the allow-list maintained by IS.

### DIFF
--- a/.github/actions/image-setup/action.yml
+++ b/.github/actions/image-setup/action.yml
@@ -12,6 +12,29 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Check connectivity for SFTP upload
+      shell: bash
+      run: |
+        set -eux
+
+        # Try to get own public IP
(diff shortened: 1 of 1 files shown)


simondeziel:
It seems like IS' list is indeed quite out of sync:

```
$ for ip in $(wget -qO - https://api.github.com/meta | jq -r '.actions | .[]' | grep -vF :); do grep -qwF "${ip}" external/auto/github.yaml || echo "${ip} is missing"; done | grep -cwF missing
389
```

Even with that finding, I'd still like to get data from failed connectivity checks just to be certain before asking IS.

PR #569, opened 2025-08-27
tests/storage-disks-vm: Add test for `lxd file push --uid` to VM

skozina:


tomponline:
@skozina you can re-run the latest/edge tests when possible because the fix just landed in latest/edge and should pass now.

tomponline (dismissed):

tests/storage-disks-vm
```diff
@@ -162,6 +162,13 @@ lxc config device remove v1 cloudinit
 ! lxc exec v1 -- test -e "${BLOCK_SPECIAL_FILE}" || false
 ! lxc exec v1 -- test -e /dev/disk/by-id/scsi-0QEMU_QEMU_CD-ROM_lxd_cloudinit || false
 
+# Test file pushes to VMs with uid and gid changes
+touch "${testRoot}/file-push.txt"
+lxc file push "${testRoot}/file-push.txt" v1/root/file-push.txt --uid=65535 --gid=65535
+[ "$(lxc exec v1 -- stat -c "%u" /root/file-push.txt)" = "65535" ]
+[ "$(lxc exec v1 -- stat -c "%g" /root/file-push.txt)" = "65535" ]
```
simondeziel, line 169: Exec'ing in a VM isn't particularly fast so we other places in the script testing owership do both the UID and GID in once go:

```suggestion
[ "$(lxc exec v1 -- stat -c "%u:%g" /root/file-push.txt)" = "65535:65535" ]
```

simondeziel (approved):

tests/storage-disks-vm
skozina: Thanks for the tip! Applied..

PR #570, opened 2025-08-27
tests/storage-disks-vm: Extend the test for `lxd file push --uid` to include directory push

skozina:


simondeziel (approved):
LGTM and works in `latest/edge`, thanks!

skozina:
> LGTM and works in `latest/edge`, thanks!

Thanks for testing;-)

skozina:
This is expected to pass on all releases.
The fix for https://github.com/canonical/lxd/pull/16335 only fixing the behavior on containers. This lxd-ci PR only adds similar regression tests for VMs, but these were not broken, and we're not fixing them;-) The failing tests don't seem to be related.

skozina:
The https://github.com/canonical/lxd/pull/16335 PR contains the necessary tests for the container case, which were broken (but missing, so unnoticed), so the PR provides both the fix and the respective tests for containers.

tomponline (approved):
Got it thanks

PR #582, opened 2025-09-17
tests/gpu-container: Only test CDI mode on LXD servers with CDI mode support

tomponline:
Part of 5.0.5 testing


tomponline:
> Small nit, but I would just add one output to indicate the test was skipped

that doesn't seem to be the convention in this repo?

PR #584, opened 2025-09-17
bin/test-image: fix logic to detect unpriv containers for Debian workaround

simondeziel:


tomponline (dismissed):
what is security.privileged=false

PR #591, opened 2025-10-02
tests/container: check running misc bins with binfmt_misc

kadinsayani:
This pull request adds tests for `binfmt_misc` namespace support in unprivileged containers. `binfmt_misc` enables registering interpreters for binary types, allowing for execution of miscellaenous binaries. For this test, we enable `binfmt_misc`, add `qemu-user` binary types (including `qemu-aarch64`), compile a simple C program for `arm64`, and execute it in an `amd64` container.

I considered a few different approaches for testing `binfmt_misc`, including downloading the `arm64` variant of the `hello` package or downloading a statically compiled `arm64` busybox binary. Compiling a C program seemed to be the most simple approach since we don't need to add new package sources or install binaries from sources with URLs subject to change. Additionally, this approach covers a common use-case for the feature.

Depends on https://github.com/canonical/lxd/pull/16848.

### DIFF
--- a/tests/container
+++ b/tests/container
@@ -214,6 +214,7 @@ if ! { runsMinimumKernel 6.9 && hasNeededAPIExtension container_bpf_delegation; 
         lxc init "${IMAGE}" c1
         lxc config set c1 security.delegate_bpf=true
         ! lxc start c1 || false
+        lxc config unset c1 security.delegate_bpf
     fi
 
     echo "Skipping unprivileged container bpf delegation test as the kernel is too old or LXD support missing"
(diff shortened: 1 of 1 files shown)


tests/container
```diff
@@ -233,6 +233,32 @@ else
     lxc exec c1 -- grep -F "delegate_cmds=map_create:prog_load,delegate_maps=ringbuf,delegate_progs=socket_filter,delegate_attachs=cgroup_inet_ingress" /proc/self/mountinfo
 fi
 
+echo "==> Test running miscellaneous binary formats with binfmt_misc in unprivileged container"
+if lxc info | grep -F 'unpriv_binfmt: "true"'; then
+	lxc launch "${IMAGE}" c2
+	waitInstanceBooted c2
```
simondeziel, line 239: Can we keep using `c1` that's already booted and ready? If not, I'd rather have `c1` deleted before then.

kadinsayani: Yes, we can use `c1` :)

tests/container
```diff
@@ -214,6 +214,7 @@ if ! { runsMinimumKernel 6.9 && hasNeededAPIExtension container_bpf_delegation;
         lxc init "${IMAGE}" c1
         lxc config set c1 security.delegate_bpf=true
         ! lxc start c1 || false
+        lxc config unset c1 security.delegate_bpf
```
simondeziel: Ah, I had not look carefully at what baggage `c1` had accumulated. Considering the BPF delegation `else` branch, I think you were better off with a fresh instance.

tests/container
```diff
@@ -233,6 +234,36 @@ else
...
+
+	echo "Enabling binfmt_misc in unprivileged container"
+	lxc exec c1 -- mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc
+    [ "$(lxc exec c1 -- cat /proc/sys/fs/binfmt_misc/status)" = "enabled" ]
+
+	echo "Installing required packages and compiling aarch64 test binary"
+	lxc exec c1 -- apt-get update
+	lxc exec c1 -- apt-get install --no-install-recommends -y qemu-user-static gcc-aarch64-linux-gnu binfmt-support libc6-dev-arm64-cross
+	lxc exec c1 -- systemctl restart systemd-binfmt # Restart systemd-binfmt service to pick up new binary types provided by qemu-user
+	lxc exec c1 -- test -f /proc/sys/fs/binfmt_misc/qemu-aarch64
```
simondeziel, line 253: This has more to do with verification so I'd put it right before `./hello == "Hello, World!"`

tests/container
```diff
@@ -233,6 +234,36 @@ else
...
+
+int main() {
+    printf("Hello, World!\n");
+    return 0;
+}
+EOF
+	lxc exec c1 -- aarch64-linux-gnu-gcc -static hello.c -o hello
+
+	echo "Running aarch64 test binary"
+	[ "$(lxc exec c1 -- ./hello)" = "Hello, World!" ]
```
simondeziel, line 262: I'd like some negative testing added as well, something like this:
```suggestion
	[ "$(lxc exec c1 -- ./hello)" = "Hello, World!" ]
	lxc exec c1 -- umount /proc/sys/fs/binfmt_misc
    ! lxc exec c1 -- ./hello || false
```

tests/container
```diff
@@ -231,6 +232,44 @@ else
...
+echo "==> Test running miscellaneous binary formats with binfmt_misc in unprivileged container"
+if lxc info | grep -F 'unpriv_binfmt: "true"'; then
+	lxc start c1
+	waitInstanceBooted c1
+
+	echo "Enabling binfmt_misc in unprivileged container"
+	lxc exec c1 -- mount binfmt_misc -t binfmt_misc /proc/sys/fs/binfmt_misc
+    [ "$(lxc exec c1 -- cat /proc/sys/fs/binfmt_misc/status)" = "enabled" ]
+
+	echo "Installing required packages and compiling aarch64 test binary"
```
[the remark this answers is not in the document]
simondeziel: Go is readily available on GH Action runners.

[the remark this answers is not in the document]
kadinsayani: Sweet, we can compile on the runner and simply push the binary to the container. Thanks!

tests/container
```diff
@@ -231,6 +232,52 @@ else
...
+func main() {
+    fmt.Println("Hello, World!")
+}
+EOF
+
+    GOOS=linux GOARCH=arm64 go build -o hello hello.go
+
+    echo "Pushing binary to container"
+    lxc file push hello c1/root/hello
+    lxc exec c1 -- chmod +x /root/hello
```
simondeziel: `lxc file push --quiet` here please. Also, `lxc file push` already preserves the execute bit that `go build` have set so no need to `chmod`.

PR #593, opened 2025-10-06
tests/network-ovn: Host route for LB external IPs

mkalcok:
Logical Routers in OVN 22.03 do not support answering ARP requests for Load Balancer IP addresses that are not part of the subnet on one of their Logical Router Ports[0].
This makes some of the tests in the `forward` suite flaky, because these tests set up Load Balancers with external IP addresses that don't belong to any of the router subnets. As such, they rely on occasional left-over ARP entry for the LB IP to pass.

There's already an established precedent in the network-ovn test suites to add a host route for the Load Balancer IP via the IP address of the Logical Router Port.
This change adds the host routes in test suites where it was missing.

[0] https://github.com/ovn-org/ovn/issues/124

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -641,10 +641,12 @@ ovn_forward_tests() {
     echo "==> Add routes to the uplink network and check we can create forwards"
     lxc network set lxdbr0 ipv4.routes=192.0.2.0/24 ipv6.routes=2001:db8:1:2::/64 --project=default
 
-    echo "==> Add a static route to SNAT address on uplink to OVN network's router (simulating BGP route advert to uplink)"
-    # XXX: only needed for IPv6 now (see https://github.com/ovn-org/ovn/issues/124)
+    echo "==> Add static routes to network forward external address on uplink to OVN network's router (simulating BGP route advert to uplink)"
+    # XXX: (see https://github.com/ovn-org/ovn/issues/124)
     ovnIPv6="$(lxc network get ovn-virtual-network volatile.network.ipv6.address)"
+    ovnIPv4="$(lxc network get ovn-virtual-network volatile.network.ipv4.address)"
     ip r add 2001:db8:1:2::1/128 via "${ovnIPv6}" dev lxdbr0
+    ip r add 192.0.2.1/32 via "${ovnIPv4}" dev lxdbr0
 
     echo "==> Check forwards can be created with no target address"
     lxc network forward create ovn-virtual-network 192.0.2.1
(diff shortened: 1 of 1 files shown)


simondeziel (approved):
LGTM, thanks for providing this.

On nit though. Could we lose the `|| true` with the route cleanups?

mkalcok:
The `|| true` was dropped from the route cleanup calls, and I also relocated one of the cleanup blocks because it wasn't in the intended (`test_load_balancer`) test suite.

Regarding the failing LB test from the previous run, I don't think it was caused by this change.
https://github.com/canonical/lxd-ci/blob/7e56f07c9768d7008197562a3eb5375edd3cfa66/tests/network-ovn#L1201-L1214

I managed to reproduce it locally after couple of tries, and if I understand it correctly, the test tries to do 5 iterations in order to hit different backend of the LB. However that is not guaranteed to happen. Have there been any problems occurring with this test, or was the failure indeed introduced by this change?

tomponline:
> The `|| true` was dropped from the route cleanup calls, and I also relocated one of the cleanup blocks because it wasn't in the intended (`test_load_balancer`) test suite.
> 
> Regarding the failing LB test from the previous run, I don't think it was caused by this change.
> 
> https://github.com/canonical/lxd-ci/blob/7e56f07c9768d7008197562a3eb5375edd3cfa66/tests/network-ovn#L1201-L1214
> 
> I managed to reproduce it locally after couple of tries, and if I understand it correctly, the test tries to do 5 iterations in order to hit different backend of the LB. However that is not guaranteed to happen. Have there been any problems occurring with this test, or was the failure indeed introduced by this change?

@mkalcok ah, ok, are you thinking the round-robin behaviour is not behaving the same in ovn deb in 22.04 as in the later microovn versions (which arent failing afaik)?

tomponline:
Ah ok thanks

@nmezhenskyi @simondeziel was this based on observation at the time of writing the tests or something else?

Originally discussed here: https://github.com/canonical/lxd-ci/pull/560#discussion_r2270260617

mkalcok:
Looking at [OVN system tests](https://github.com/ovn-org/ovn/blob/708707672e7c7d4a13aa53759cffff209b585d5e/tests/system-ovn.at#L1932-L1967). There doesn't seem to be a more surefire way to test that LB is actually load balancing. The only difference is that they roll the dice more times

tomponline:
Great, so we should increase the iteration count. Thanks @mkalcok

PR #629, opened 2025-11-27
tests/conversion: test `lvm` early as it often fail

simondeziel:
It's best to fail fast to not waste too much CI resources and allow for a quicker manual retry.


tomponline (approved):
Do you think this could be caused by the lvm daemon running on the GH runner interfering with the lvm calls inside the snap?

Could we try using `lvm.external` to see if this fixes it?

PR #685, opened 2026-02-14
bin/test-image: cope with `cloud-init` reconfiguring the network

simondeziel:
This fixes the Ubuntu Questing images that recently regressed.


tomponline (approved):
Ta

PR #724, opened 2026-03-30
bin/helpers: Do not set northbound_connection when using MicroOVN

nmezhenskyi:
This is needed to allow LXD to dynamically pick up the correct address for Northbound database from MicroOVN.

Follow-up to https://github.com/canonical/lxd/pull/17874.


simondeziel (dismissed):

simondeziel (approved):
Right, looks good, thanks!

tomponline (approved):
ta

PR #736, opened 2026-04-14
github: switch mirror for openEuler (again)

simondeziel:


tomponline (dismissed):

simondeziel:
With the North American mirror used throughout the build, it build the image in less than 10 minutes while previously it would often times hit the 1 hour time limit.

MusicDin (approved):

PR #737, opened 2026-04-16
tests/network-ovn: small tweaks

simondeziel:


bin/local-run
```diff
@@ -5,6 +5,10 @@ set -ux
 if command -v cloud-init > /dev/null; then
     echo "Waiting for cloud-init"
     cloud-init status --wait --long
+
+    # cloud-init might put environment variables which we'd miss if we 'lxc shell'ed in too early
+    . /etc/environment
```
tomponline, line 10: How did this come up?

tomponline (approved):

bin/local-run
simondeziel: If you `lxc launch` your ephemeral VM and `lxc shell` into it before cloud-init has tweaked `/etc/environment`, you miss the `PURGE_LXD=1` added by cloud-init.

tomponline: Thanks, how did you come across this? Was it causing failing tests?

simondeziel: Oh no, it's just to reduce **user** friction. CI is entirely unaffected as machines are patient, not humans ;)

PR #742, opened 2026-04-22
Refine workaround for VM inside ctn

simondeziel:
I verified the Questing `-proposed` kernel and it indeed addresses https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2147374

The kernel still needs to travel some more before it reaches us:
* `questing-proposed` -> `questing-updates`
* Go through HWE backports so that it lands in `noble-updates`
* GHA runner images are updated to pick up the `-24-azure` variant

But with this PR, our tests will automatically re-enable the needed bits to exercise the fixed code paths.

### DIFF
--- a/tests/storage-metadata
+++ b/tests/storage-metadata
@@ -1,9 +1,10 @@
 #!/bin/bash
 set -eux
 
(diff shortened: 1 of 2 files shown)


simondeziel:
Sigh, `-azure` use a much different build number so that `uname -r` looks like `6.17.0-1010-azure`

PR #745, opened 2026-04-24
tests: Account for `restricted.registries` project config option

nmezhenskyi:
This PR updates tests that are using restricted projects to account for the new `restricted.registries` config option.

It updates the `network-ovn` and `storage-disks-vm` to set `restricted.registries=allow` for their restricted projects. This ensures that instances can be created in these tests.

### DIFF
--- a/tests/network-ovn
+++ b/tests/network-ovn
@@ -157,6 +157,11 @@ ovn_basic_tests() {
         -c features.networks=true \
         -c restricted=true
 
+    # Allow the project to access image registries.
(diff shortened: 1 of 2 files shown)


simondeziel (approved):

simondeziel:
Thanks to the extension check, this should be safe to merge now even if the LXD side hasn't land... unless the API name is expected to change.

nmezhenskyi:
The API extension name will remain the same

PR #748, opened 2026-04-27
Tests: Accommodate OVN load balancer changes

roosterfish:
Starting with the `network_load_balancer_pool` extension a port can either specify a list of backends or a pool.

Should be merged together with https://github.com/canonical/lxd/pull/18179.


tomponline (changes requested):
where possibly please use API extension detection rather than version number detection, that way we can still test candidate channels before the specific release is made.

roosterfish:
> where possibly please use API extension detection rather than version number detection, that way we can still test candidate channels before the specific release is made.

Yes sure, I have just copied the same pattern from the previous change but now this is due to a new feature which introduces an API extension so I'll use this one instead. Thanks.

tomponline:
@roosterfish once this is updated and passing ill merge so we can test it on your new load balancers PR

tomponline (approved):
ta

PR #749, opened 2026-04-27
github: build `amd64` and `arm64` images on the same schedule trigger

simondeziel:
Previously, with 2 `cron` schedule meant the same workflow file was used to build `amd64` first and later `arm64`.

The README as a single link per image workflow which means the last image build to fire (`arm64`) was the one being visible losing visibility on `amd64`-only build/validation issues (like https://github.com/canonical/lxd/issues/18187).


tomponline (approved):
Ta

PR #755, opened 2026-04-30
bin/helpers: Ignore dnsmasq and io_context_pool crashes

tomponline:


bin/helpers
```diff
@@ -459,6 +459,18 @@ cleanup() {
         rm /var/crash/*waagent*
     fi
 
+    # Ignore io_context_pool crashes.
+    if compgen -G "/var/crash/io_context_pool*" > /dev/null 2>&1; then
+        echo "::notice::==> CORE: io_context_pool core dump ignored"
+        rm /var/crash/io_context_pool*
+    fi
```
simondeziel, line 466: Is that from `librbd` as used by QEMU? I'm not familiar with it but a quick search lead to:
> In QEMU, io_context_pool is not a core QEMU component, but rather a background thread pool spawned by librbd (the Ceph RADOS block device library) when QEMU connects to a Ceph/RBD storage backend. These threads asynchronously handle I/O requests and messaging for the Ceph cluster.

simondeziel (approved):
Thank you

bin/helpers
tomponline: im not sure

PR #757, opened 2026-05-06
github: abort if the preflight check in `actions/image-setup` fails

simondeziel:


simondeziel:
@tomponline I just compared IS' firewall list with the current one from GH and IS is only missing a few `/32` after catching up earlier today. The clustered upload failures we saw were just a hiccup.

tomponline (approved):

PR #785, opened 2026-06-12
github: use Testflinger creds from GH secrets

simondeziel:


tomponline (approved):
Thanks!

simondeziel:
FYI, I cannot see the new parameters in the job logs but maybe they are scrubbed due to using secrets. I'll keep that in mind if it fails in the future.

PR #792, opened 2026-06-19
Allow collecting coverage info from `tests/conversion`

simondeziel:


bin/helpers
```diff
@@ -451,6 +451,51 @@ teardown_swap() {
...
 
+# Returns true if Go coverage collection is enabled (GOCOVERDIR is set).
+coverage_enabled() {
+  [ -n "${GOCOVERDIR:-}" ]
+}
+
+# Setup instance to collect coverage data from running Go binaries from inside the instance.
+# If coverage is not enabled, this is a no-op.
+# If the instance is a VM, the lxd-agent will be instrumented too.
+# Note: incompatible with `migration.stateful=true` instances.
```
tomponline, line 462: why?

tomponline (approved):
ta

bin/helpers
simondeziel: That code is copy-pasted from the LXD repo. I should probably have mentioned it.

As for why, I think it because LXD doesn't authorized host shared directories when `migration.stateful=true`.

tomponline: good to include that info ta

simondeziel: https://github.com/canonical/lxd-ci/pull/794

PR #793, opened 2026-06-22
Add `ubuntu-26.04` to the testing matrix

simondeziel:
And minor test tweaks.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,13 +10,16 @@ on:
   workflow_dispatch:
     inputs:
       ubuntu-releases:
-        description: Ubuntu releases to run the tests against. In JSON format, i.e. '["22.04", "24.04"]'.
+        description: Ubuntu releases to run the tests against. In JSON format, i.e. '["22.04", "24.04", "26.04"]'.
         type: choice
-        default: '["22.04", "24.04"]'
+        default: '["22.04", "24.04", "26.04"]'
         options:
+          - '["22.04", "24.04", "26.04"]'
+          - '["24.04", "26.04"]'
           - '["22.04", "24.04"]'
           - '["22.04"]'
           - '["24.04"]'
+          - '["26.04"]'
       snap-track:
         description: Snap track to run the tests i.e. '["latest/edge", "5.0/candidate"]'.
         type: choice
@@ -68,7 +71,7 @@ jobs:
   code-tests:
     if: ${{ github.event_name == 'pull_request' || github.repository == 'canonical/lxd-ci' }}
     name: Code
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-26.04
     steps:
       - name: Checkout
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -66,9 +69,9 @@ defaults:
 
 jobs:
   code-tests:
-    if: ${{ github.event_name != 'schedule' || github.repository == 'canonical/lxd-ci' }}
+    if: ${{ github.event_name == 'pull_request' || github.repository == 'canonical/lxd-ci' }}
     name: Code
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-26.04
```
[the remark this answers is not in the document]
simondeziel: It is intentional.

tomponline:
> In order to avoid exceeding the maximum matrix size (256 jobs), stop testing `6/candidate` and `6/edge` on 22.04.

we can still manually request though right?

simondeziel:
> > In order to avoid exceeding the maximum matrix size (256 jobs), stop testing `6/candidate` and `6/edge` on 22.04.
> 
> we can still manually request though right?

No as the exclusion would remove those combinations.

simondeziel:
We are currently sitting at 222 jobs, if you want to test 22.04 with `6/*`, are there a few tests you would like more than others?
We still have 34 jobs we can fill :)

tomponline:
> We are currently sitting at 222 jobs, if you want to test 22.04 with `6/*`, are there a few tests you would like more than others? We still have 34 jobs we can fill :)

When i manually kick off a test i can specify the OS version to test on, can we have 22.04 available just for manual jobs?

simondeziel:
> > We are currently sitting at 222 jobs, if you want to test 22.04 with `6/*`, are there a few tests you would like more than others? We still have 34 jobs we can fill :)
> 
> When i manually kick off a test i can specify the OS version to test on, can we have 22.04 available just for manual jobs?

Turns out that I went too far with exclusions as it was enough to disable `4.0/*` and `5.0/*` on 26.04 to avoid hitting the limit. With that, let me reintroduce the `latest/*` testing on 22.04 by default as we are then at 252 jobs :)

.github/workflows/tests.yml
```diff
@@ -307,6 +307,19 @@ jobs:
             track: "5.21/candidate"
           - test: qemu-external-vm
             track: "5.21/edge"
+          # known broken
```
tomponline: @simondeziel please can you elaborate here? They should be fixed rather than masked no?

tomponline: I think we should not mask this but instead create a GH issue for the UI team to fix.

tomponline: Especially as we're coming up to a point release for LXD 5.21 anyway

simondeziel: Issue reported https://github.com/canonical/lxd-ui/issues/2094
I've added this link to the comment.

tomponline:
@simondeziel 

>  + '[' -n udisksd.361.gz ']'
+ echo 'CORE: coredump found during cleanup, forcing failure'
CORE: coredump found during cleanup, forcing failure
Test failed

https://github.com/canonical/lxd-ci/actions/runs/28094957235/job/83182063285?pr=793

tomponline (approved):

simondeziel:
> @simondeziel
> 
> > * '[' -n udisksd.361.gz ']'
> 
>     * echo 'CORE: coredump found during cleanup, forcing failure'
>       CORE: coredump found during cleanup, forcing failure
>       Test failed
> 
> 
> https://github.com/canonical/lxd-ci/actions/runs/28094957235/job/83182063285?pr=793

Interesting, this is `udisk2` from an instance that's crashing because since https://github.com/canonical/lxd/commit/e869d33aee842245256a0944e7efafcf1befd2d4, we remove it from the host.

I'll see what I can do with this one.

PR #794, opened 2026-06-22
6.17 kernels now have the fix for LP: #2147374

simondeziel:
Also add a note  explaining why `setup_instance_gocoverage` doesn't work with `migration.stateful=true`


tomponline (approved):
nice

PR #800, opened 2026-06-25
Test improvements

simondeziel:


simondeziel:
@tomponline the only failure is in the LXD UI tests on `5.21/edge` and the UI team is aware.

tests/storage-disks-vm
```diff
@@ -268,6 +271,7 @@ DEVICE_HOTPLUG="1"
 echo "==> Check /dev/disk/by-id with escaped long device name"
 lxc launch "${IMAGE}" v2 --vm
 waitInstanceReady v2
+#setup_instance_gocoverage v2 # The injected disk device would not be allowed in this restricted project
```
tomponline, line 274: these commented lines should be removed?

simondeziel: I added them to highlight that those spots were not forgotten about, to document why it would not apply. I don't feel strongly either way.

tomponline: I see thanks

tomponline (approved):

PR #801, opened 2026-06-29
Ubuntu Questing (25.10) is EOL

simondeziel:


tomponline (approved):

simondeziel:
I just learned that 25.10 is still supported [until July 9th](https://lists.ubuntu.com/archives/ubuntu-announce/2026-June/000324.html).

I initially based the EOL date on the [announcement](https://discourse.ubuntu.com/t/questing-quokka-release-notes/59220):
> Ubuntu 25.10 will be supported for 9 months until July 2026.

And it seems like https://endoflife.date/ubuntu was similarly confused.

Either way, the last published images are kept for ~14 days on `images:` so it shouldn't be too bad.

PR #802, opened 2026-06-29
tests/gpu-container: Use API extension guards for legacy nvidia runtime

nmezhenskyi:
This accounts for the incoming removal of `nvidia.*` instance-level config options from LXD `6.10` https://github.com/canonical/lxd/pull/18559.

### DIFF
--- a/tests/gpu-container
+++ b/tests/gpu-container
@@ -88,14 +88,16 @@ lxc config device add c1 gpus gpu
 sleep 1
 [ "$(lxc exec c1 -- ls /dev/dri/ | grep -c '^card[0-9]')" = "${total_gpu}" ]
 
-# Test nvidia runtime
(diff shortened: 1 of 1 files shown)


tests/gpu-container
```diff
@@ -136,7 +146,7 @@ if hasNeededAPIExtension "gpu_cdi"; then
 
     lxc stop --force c1
 
-    lxc config set c1 security.nesting=true nvidia.driver.capabilities=all
+    lxc config set c1 security.nesting=true
```
nmezhenskyi, line 149: This was actually unnecessary and was simply ignored because `nvidia.driver.capabilities` requires the `nvidia.runtime=true`. So it's safe to remove it from the CDI section here.

simondeziel (approved):

tomponline (approved):

PR #805, opened 2026-06-30
bin/local-run: only wait for cloud-init in VMs

simondeziel:
Some Testflinger machines are unable to get a successful `cloud-init status --wait` which prevents them reusing those CI tests.

This wait was added to ensure a local test VM using the `lxd-ci` profile was fully ready before launching the test script.

By waiting only when a VM is detected, both should be happy.

### DIFF
--- a/bin/local-run
+++ b/bin/local-run
@@ -2,8 +2,8 @@
 set -eux
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


nmezhenskyi (approved):
Thanks!

PR #808, opened 2026-07-03
tests/cpu-vm: poll guest CPU count to improve reliability

simondeziel:


simondeziel:
@tomponline lots of red due to the snapstore having a rough day but the `cpu-vm` test that's changed here is passing all around.

kadinsayani (approved):

PR #819, opened 2026-07-22
tests/ui: align UI tests with UI bundled in the snap

simondeziel:


simondeziel:
@tomponline I told you the `source-commit` was saved even for unpinned parts but it turned out to be wrong :/
This means the pinning down to the commit will only work on tests against candidate and stable snaps which IMHO greatly reduces the value of this PR.

Your call if you still want it or not.

tomponline:
> This means the pinning down to the commit will only work on tests against candidate and stable snaps which IMHO greatly reduces the value of this PR.
> 
> Your call if you still want it or not.

Absolutely I think its of value. I didn't expect the edge builds to be able to use pinned tests.

tomponline (approved):
Thanks!

PR #835, opened 2026-08-18
tests/network-bridge-firewall: refresh apt package definitions

simondeziel:


roosterfish (approved):

tomponline:
@simondeziel is there something we can do about:

```
snap set lxd minio.path=/opt/minio
+ systemctl reload snap.lxd.daemon
+ lxd waitready --timeout=300
Error: LXD daemon is shutting down after 300s timeout
```

Seems to be causing failures on several tests.

Ta

PR #837, opened 2026-08-18
tests: Add snapcraft integration test

nmezhenskyi:
This PR adds a dedicated test for verifying that LXD and Snapcraft work correctly together.

This is helpful for making sure that the incoming image registries LXD feature does not break anything for Snapcraft.

### DIFF
--- a/tests/snapcraft
+++ b/tests/snapcraft
@@ -1,0 +1,63 @@
+#!/bin/bash
+set -eux
+
+# Install and setup LXD.
+install_lxd
+lxd init --auto
+
+# Install snapcraft (use latest/candidate to help catch regressions in LXD/snapcraft earlier).
+snap install snapcraft --classic --channel=latest/candidate
+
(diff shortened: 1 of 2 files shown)


tests/snapcraft
```diff
@@ -0,0 +1,60 @@
...
+parts:
+  example-script:
+    plugin: dump
+    source: src/
+    organize:
+      example.sh: bin/example.sh
+EOF
+
+# Produce the snap file.
+snapcraft pack -v
```
simondeziel, line 50: Considering that `snapcraft` support various modes of building should we supplement this with a `lxc list --project snapcraft` or similar to ensure the LXD driver was used.

simondeziel (commented):
LGTM with a tiny suggestion.

simondeziel (approved):

tests/snapcraft
nmezhenskyi: Updated.

PR #839, opened 2026-08-18
tests: Add lxd-consumers integration test

nmezhenskyi:
This PR renames `tests/snapcraft` into `tests/lxd-consumers` and adds a dedicated test for verifying that LXD and Workshop work correctly together.  

This is helpful for making sure that the incoming image registries LXD feature does not break anything for Workshop.


nmezhenskyi:
@simondeziel It seems this PR exceeds the allowed matrix limit of `256` jobs per workflow, and as a result no jobs get dispatched.

So it's probably not worth it to add this test to the workflow.

Maybe I should just add the `tests/workshop` script to be able to run it locally, but do not add it to matrix?

tests/workshop
```diff
@@ -0,0 +1,36 @@
...
+workshop init dev --base ubuntu@24.04
+
+# The definition should now exist in the project directory.
+[ -f "${tmpdir}/.workshop/dev.yaml" ]
+
+# Launch the workshop (downloads the base image and provisions the LXD container).
+workshop launch dev
+
+# Verify the workshop is usable by running a command inside it and checking the base.
+workshop exec dev -- cat /etc/os-release | grep -qF 'VERSION_ID="24.04"'
```
simondeziel, line 25:
```suggestion
workshop exec dev -- grep -xF 'VERSION_ID="24.04"' /etc/os-release
```

tests/workshop
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -eux
+
+# Install and setup LXD.
+install_lxd
+lxd init --auto
+
+# Install Workshop.
+snap install workshop --classic --channel=latest/stable
```
simondeziel, line 9: Worth using the candidate channel?

simondeziel (commented):
To limit the number of concurrent jobs, could we have a single script verifying that LXD works for all Canonical products? Besides `snapcraft` and `workshop`, are there other tools to test like `charmcraft`?

Dunno what'd be a good name? `tests/lxd-substrate`, `tests/crafts`, `tests/canonical`?

Then that script could accept an optional list of Canonical products to test with a default of all of them?

tests/workshop
nmezhenskyi: Workshop only has `stable` and `edge`. Should we go with `edge`?

simondeziel: Yeah, let's go with `edge`, please.

tomponline:
> To limit the number of concurrent jobs, could we have a single script verifying that LXD works for all Canonical products? Besides `snapcraft` and `workshop`, are there other tools to test like `charmcraft`?
> 
> Dunno what'd be a good name? `tests/lxd-substrate`, `tests/crafts`, `tests/canonical`?
> 
> Then that script could accept an optional list of Canonical products to test with a default of all of them?

Agree. How about lxd-consumers?

nmezhenskyi:
> Agree. How about lxd-consumers?

I like this! I'll go with `tests/lxd-consumers`.

tests/workshop
nmezhenskyi: Updated.

tests/workshop
nmezhenskyi: Updated.

tests/lxd-consumers
```diff
@@ -0,0 +1,118 @@
...
+    workshop remove dev
+    cd - > /dev/null
+    rm -rf "${tmpdir}"
+}
+
+# Install and setup LXD.
+install_lxd
+lxd init --auto
+
+for product in ${PRODUCTS}; do
```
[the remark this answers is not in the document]
nmezhenskyi: Fixed.

tests/lxd-consumers
```diff
@@ -0,0 +1,180 @@
+#!/bin/bash
+set -eux
+
+# Integration test verifying that Canonical products relying on LXD keep working
+# with it. It accepts an optional list of products to test as arguments.
+# When no arguments are provided, all supported products are tested.
+if [ "${#}" -gt 0 ]; then
```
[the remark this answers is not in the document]
nmezhenskyi: Unnecessary.

simondeziel (approved):
Nicely done!

PR #843, opened 2026-08-21
tests/devlxd: merged `devlxd-container` and `devlxd-vm`

simondeziel:
With this, the `system-tests` matrix goes from 256 jobs to 249 jobs.


tests/devlxd
```diff
@@ -0,0 +1,277 @@
...
+# Allow for running a specific set of tests.
+if [ "$#" -gt 0 ]; then
+    for test_name in "$@"; do
+        case "${test_name}" in
+            container|devlxd-container|test_container|test_devlxd_container)
+                test_devlxd_container
+                ;;
+            vm|devlxd-vm|test_vm|test_devlxd_vm)
+                test_devlxd_vm
+                ;;
```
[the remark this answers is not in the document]
simondeziel: Those calling scripts are effectively dead ones that will soon be removed.

.github/workflows/tests.yml
```diff
@@ -178,8 +177,6 @@ jobs:
             track: "4.0/candidate"
           - test: cpu-vm
             track: "4.0/candidate"
-          - test: devlxd-vm
-            track: "4.0/candidate"
```
yinbin-dev, line 182: Previously `devlxd-vm` tests were excluded from `4.0/candidate`, does this change preserve this behavior or we want to include them now?

.github/workflows/tests.yml
```diff
@@ -234,8 +231,6 @@ jobs:
             track: "4.0/edge"
           - test: cpu-vm
             track: "4.0/edge"
-          - test: devlxd-vm
```
yinbin-dev, line 237: Previously `devlxd-vm` tests were excluded from `4.0/edge`, does this change preserve this behavior or we want to include them now?

yinbin-dev (commented):
LGTM, asked a question for my understanding though.

tomponline (approved):
Thanks!

.github/workflows/tests.yml
simondeziel: The exclusion was moved into `test_devlxd_vm` itself and it skips VM tests for all `4.0/*`.

.github/workflows/tests.yml
simondeziel: The exclusion was moved into `test_devlxd_vm` itself and it skips VM tests for all `4.0/*`.

PR #844, opened 2026-08-21
github: Improve GPU passthrough test matrix

nmezhenskyi:
This PR excludes AMD tests from 5.21 LXD series.

It also fixes previous AMD and CDI jobs exclusions using the `matrix.exclude`, so that these jobs are only skipped where necessary when using multi-track dispatch.

### DIFF
--- a/.github/workflows/gpu-passthrough-tests.yml
+++ b/.github/workflows/gpu-passthrough-tests.yml
@@ -49,11 +49,14 @@ env:
 jobs:
   uc-nvidia-cdi:
     runs-on: [self-hosted, self-hosted-linux-amd64-noble-private-endpoint-small]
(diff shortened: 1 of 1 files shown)


tomponline (approved):
ta

tomponline:
@nmezhenskyi doesn't look quite right still when running on 5.21/candidate channel:

https://github.com/canonical/lxd-ci/actions/runs/32495143021/job/96811688976

nmezhenskyi:
@tomponline Yes, just noticed this. Preparing a fix. Will revert to the `if` approach since we don't really use multi-track dispatches at once anyway.

The issue with `matrix.exclude` is it only works when there is actually a multi-track dispatch that has at least one included track 😔

PR #845, opened 2026-08-21
github: Fix GPU passthrough test matrix

nmezhenskyi:
Follow-up to the https://github.com/canonical/lxd-ci/pull/844.

Fixes the issue where jobs are not correctly excluded when dispatched using a single track.


tomponline (approved):
ta

PR #847, opened 2026-08-25
bin/helpers: wait for lxd-agent to be consistently responsive to execs

jrtorsella:
This works around a quirk in the handling of an lxd-agent restart during which a single agent goes from being able to service execs to being masked by a delayed STOPPED flag to being able to service them again. See the code comment for details on the scenario this is working around, but essentially:

lxc exec systemctl restart lxd-agent.service
first lxc-agent.service goes down for the count, STOPPED state is not necessarily observed by lxd because observations follow QEMU channel signal cadence and this is rate-limited to once a second.
lxc exec can succeed in this window, it seems to hang and is then picked up by the new lxd-agent
State of STOPPED observed by lxd at some point after the above lxc exec gets through
Until the next rate-limited qemu event, lxc execs now fail

The test fix here is to wait for two consecutive execs to succeed 1s apart from each other.

### DIFF
--- a/bin/helpers
+++ b/bin/helpers
@@ -536,12 +536,35 @@ EOF
   lxc exec "${instance}" --project "${project}" -- systemctl restart --no-block lxd-agent.service || true
 
   for _ in $(seq 60); do
-      rc=0
-      start_time_new=$(lxc exec "${instance}" --project "${project}" -- systemctl show -p ExecMainStartTimestampMonotonic --value lxd-agent.service) || rc=$?
-      if [ "${rc}" -eq 0 ] && [ -n "${start_time_new}" ] && [ "${start_time_new}" != "${start_time_orig}" ]; then
-        return 0
-      fi
-      sleep 1
+    rc=0
+    start_time_new=$(lxc exec "${instance}" --project "${project}" -- systemctl show -p ExecMainStartTimestampMonotonic --value lxd-agent.service) || rc=$?
+    if [ "${rc}" -eq 0 ] && [ -n "${start_time_new}" ] && [ "${start_time_new}" != "${start_time_orig}" ]; then
+      local consecutive=0
+      for _ in $(seq 30); do
+        # Check that the instance can serve 'lxc exec' reliably. Test failures revealed that despite being
+        # able to handle the above lxc exec to populate start_time_new, the agent can flip back to appearing
+        # offline (such that an lxc exec will fail) due to some quirks in the online/offline status handling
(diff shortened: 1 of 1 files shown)


bin/helpers
```diff
@@ -536,12 +536,35 @@ EOF
...
-      fi
-      sleep 1
+    rc=0
+    start_time_new=$(lxc exec "${instance}" --project "${project}" -- systemctl show -p ExecMainStartTimestampMonotonic --value lxd-agent.service) || rc=$?
+    if [ "${rc}" -eq 0 ] && [ -n "${start_time_new}" ] && [ "${start_time_new}" != "${start_time_orig}" ]; then
+      local consecutive=0
+      for _ in $(seq 30); do
+        # Check that the instance can serve 'lxc exec' reliably. Test failures revealed that despite being
+        # able to handle the above lxc exec to populate start_time_new, the agent can flip back to appearing
+        # offline (such that an lxc exec will fail) due to some quirks in the online/offline status handling
```
tomponline: Online/offline?

jrtorsella: STARTED/STOPPED in the code, or 'able to service lxc exec operations vs unable to service them'.

bin/helpers
```diff
@@ -536,12 +536,35 @@ EOF
...
 
   for _ in $(seq 60); do
-      rc=0
-      start_time_new=$(lxc exec "${instance}" --project "${project}" -- systemctl show -p ExecMainStartTimestampMonotonic --value lxd-agent.service) || rc=$?
-      if [ "${rc}" -eq 0 ] && [ -n "${start_time_new}" ] && [ "${start_time_new}" != "${start_time_orig}" ]; then
-        return 0
-      fi
-      sleep 1
+    rc=0
+    start_time_new=$(lxc exec "${instance}" --project "${project}" -- systemctl show -p ExecMainStartTimestampMonotonic --value lxd-agent.service) || rc=$?
```
tomponline: What is it ur trying to do here?

Ensure the the old lxd-agent has stopped and the new one is responding to commands?

Any reason we cant use the pid changing to indicate the new one has started and is running commands rather than time?

jrtorsella: We could use the pid, sure, but it would be just as complicated - still a systemctl show, and it would be vulnerable to pid reuse. ExecMainStartTimestampMonotonic is systemd's version of stat starttime.

This part hasn't changed since the last merge - I just realized that I'd gotten the indent wrong initially and decided to fix it.

However we track "a new agent is started", the complicated part of this is the newer middle part.

tomponline (approved):
Ta!

PR #848, opened 2026-08-26
LXD-4910(tests): New snap-shutdown-order test to validate systemd shutdown ordering

jrtorsella:
This contains the new snap-shutdown-order test which is enabled only for V6 until backport. The new test verifies that LXD is shut down prior to any microceph/microovn service being shut down.

Depends on: https://github.com/canonical/lxd/pull/18926

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -131,6 +131,7 @@ jobs:
           - network-routed
           - pylxd
           - qemu-external-vm
+          - snap-shutdown-order
           - snapcraft
           - snapd
           - storage-buckets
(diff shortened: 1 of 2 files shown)


tomponline:
@jrtorsella agree, please can you update this PR description to link to the PR it depends on. Helps reviewers to order merges.

jrtorsella:
@simondeziel I've become aware since I took out this PR that we are running up against some sort of limit for tests - assuming that means it would be better for me to merge this small one into something else, is that right?

simondeziel:
@jrtorsella yes indeed we are hitting the job limit. Hopefully, that shouldn't be an issue once `lxd-ci` tests are merged into LXD as there they won't be run against all the snap versions at once.

tomponline:
Please can this be moved to lxd repo now. Cc @simondeziel for more info.

